PR #2, opened 2024-01-14
Refactoring

vedran-kasalica:
The PR aims to
- cleanup the demo example,
- document the preprocessing steps,
- introduce constraints for channel names (this will be further improved later),
- removed some unused elements from `PreprocessingPipeline` class and renamed it to `PreprocessingPipelineConfig`. The next step could be to remove the class entirely, but I didn't want to make such a big change last minute.


vedran-kasalica:
The changes look good, this could be merged as far as I'm concerned.

Erikpostt (approved):
Looking good from my side, only issue seems to be a dependency error in Poetry! Could you solve that?

vedran-kasalica:
> Looking good from my side, only issue seems to be a dependency error in Poetry! Could you solve that?

I just saw your comment, that is really strange, because tsdf version 0.5.2 is the latest version on PyPi. I will look into it

PR #6, opened 2024-02-16
Fix tsdf dependency

vedran-kasalica:
Resolving the issue with the tsdf v0.5.2 dependency found in PR https://github.com/biomarkersParkinson/dbpd-toolbox/pull/2.

In addition, the first tests are added, so the CI should pass.


Erikpostt (approved):
Looking good, merging.

PR #12, opened 2024-03-14
Predictions

Erikpostt:
Note: predictions are all False, probably has something to do with the orientation. Don't mind this for now, as long as it all functions well we should be happy.


vedran-kasalica:
Hi @Erikpostt, thank you for making the PR, we will have a look at it!

vedran-kasalica (approved):
Issues that should be addressed after the PR merge: https://github.com/biomarkersParkinson/dbpd-toolbox/issues/16 and https://github.com/biomarkersParkinson/dbpd-toolbox/issues/17

PR #13, opened 2024-03-15
Ppg toolbox deliverables until predictions

KarsVeldkamp:
Hi Vedran,

I tried to create a PPG branch in similar fashion as Erik did (until classification). I also propose some restructuring for the different folders but let me know what you think about this.

Best,
Kars


vedran-kasalica:
Thank you for creating the PR, I like the file structure of the code as well as the test data. It is good to separate the processed data for now, before we can run tests to compare whether they are equivalent (e.g., `acceleration_meta.json` in PPG and Gait).

We will get back to you with more questions once we have time to look into it in more detail.

PR #14, opened 2024-03-21
Gait quantification

Erikpostt:
Hey Vedran,

This PR started where #12 left off. I would therefore recommend wrapping that one up first before moving on to this one. This one is only due mid April, but I'm on holidays April 3rd - 11th.

This is not much work so far, but might become a little more advanced once everything is functioning on the PPP dataset. This all depends on whether we want to include visualization functionalities to the toolbox. We'll probably be discussing this later if it is relevant.

Cheers!


vedran-kasalica:
@Erikpostt thank you for sharing the code! I am currently looking at refactoring the current codebase and merging Kars' PR (so we can compare the raw binaries). I will look into this one and #12 after that.

vedran-kasalica (approved):
To resolve after the PR merge:
https://github.com/biomarkersParkinson/dbpd-toolbox/issues/18

PR #15, opened 2024-03-26
Notebook tests and a small gait refactor

vedran-kasalica:
Introduce Notebook tests using `papermill`

- Introduce `papermill` dependency
- Setup tests for the notebooks


vedran-kasalica:
I made some slight changes and uploaded notebook tests. @Erikpostt could you test `4.extract_features_arm_swing`? It is failing and I am not sure why.

Erikpostt:
> I made some slight changes and uploaded notebook tests. @Erikpostt could you test `4.extract_features_arm_swing`? It is failing and I am not sure why.

Hi @vedran-kasalica, I think I resolved the issue in the recent commit, but please verfiy. It had to do with channel names, which I've now standardized to accelerometer/gyroscope instead of acceleration/rotation, which I think it a more accurate and reasonable representation of the channels. This also modified the test data.

vedran-kasalica:
> > I made some slight changes and uploaded notebook tests. @Erikpostt could you test `4.extract_features_arm_swing`? It is failing and I am not sure why.
> 
> Hi @vedran-kasalica, I think I resolved the issue in the recent commit, but please verfiy. It had to do with channel names, which I've now standardized to accelerometer/gyroscope instead of acceleration/rotation, which I think it a more accurate and reasonable representation of the channels. This also modified the test data.

Thanks, @Erikpostt, it runs fine now, including the papermill tests.

vedran-kasalica:
@Erikpostt could you also confirm whether Notebook 5 runs for you?

Erikpostt:
> @Erikpostt could you also confirm whether Notebook 5 runs for you?

Should be fixed now.

vedran-kasalica:
> > @Erikpostt could you also confirm whether Notebook 5 runs for you?
> 
> Should be fixed now.

Thanks, it works now!

tests/test_notebooks.py
```diff
@@ -0,0 +1,245 @@
...
+        ("accelerometer_meta.json", "accelerometer_time.bin"),
+    ]
+gyroscope_binaries_pairs: list[tuple[str, str]] = [
+        ("gyroscope_meta.json", "gyroscope_samples.bin"),
+        ("gyroscope_meta.json", "gyroscope_time.bin"),
+    ]
+imu_binaries_pairs: list[tuple[str, str]] = accelerometer_binaries_pairs + gyroscope_binaries_pairs
+
+
+def create_tmp_folder_if_not_exists(base_path: str):
```
kretep, line 43: Why is this used when always using shared_datadir? Isn't the shared_datadir always present?

tests/test_notebooks.py
```diff
@@ -0,0 +1,245 @@
...
+
+def execute_notebook(
+    shared_datadir, notebook_name: str, input_dir: str, output_dir: str
+):
+    """
+    This function is used to execute a notebook.
+
+    Parameters
+    ----------
+    shared_datadir : str
```
kretep: I think the type of shared_datadir is Path

tests/test_notebooks.py
```diff
@@ -0,0 +1,245 @@
...
+
+def compare_data(
+    shared_datadir, notebook_output_dir_name: str, biaries_pairs: list[tuple[str, str]]
+):
+    """
+    This function is used to evaluate the output of a notebook. It evaluates it by comparing the output to a reference output.
+
+    Parameters
+    ----------
+    shared_datadir : str
```
kretep: Although I don't think the shared_datadir gets injected here (it's not a top-level test function, but a method that is called), so maybe rename it to something else to avoid confusion. It could be any path. And if only called with shared_datadir as parameter, the type will be Path here also

tests/test_notebooks.py
```diff
@@ -0,0 +1,245 @@
...
+        The path to the shared data directory.
+    biaries_pairs : list[tuple[str, str]]
+        The list of pairs of metadata and binary files to compare.
+    """
+    for metadata, binary in biaries_pairs:
```
kretep: missing an n here

kretep (changes requested):
Looks good, just some minor issues in the tests for the notebooks

tests/test_notebooks.py
vedran-kasalica: Yes, but I was getting an error when the `tmp` folder inside of it was not explicitly created (that's where we store the temp Notebook outputs).

kretep: What is the reasoning behind that? I only see that the 5th notebook writes to a 'tmp' folder

vedran-kasalica: They all do, all the notebook test outputs are written to the `data/tmp` folder. Mostly to distinguish it from the original data folder and to allow easier manual comparisons when needed. I just decided to use the `tmp` folder to be able to mirror the existing Notebook output structure.

tests/test_notebooks.py
vedran-kasalica: I fixed it.

tests/test_notebooks.py
vedran-kasalica: I agree, I fixed it.

tests/test_notebooks.py
vedran-kasalica: I fixed it.

kretep (approved):

PR #23, opened 2024-03-26
Renaming units gait

Erikpostt:
Couple of edits:
- Change units and bits of tsdf storage
- Resolve issue with creating windows/segments
- Change pathing to _gait_ subdirectory


vedran-kasalica (approved):
The changes look good!

PR #26, opened 2024-04-15
Adding functions

KarsVeldkamp:
I have updated the pipeline until HR quantification. For the HR estimation there is no 30s epoch of high quality PPG data since the demo data is very short. However, I changed the parameter to 10 s and now there is a single HR estimation possible.


vedran-kasalica (approved):
@KarsVeldkamp thank you for submitting the code. I will merge the code into the main. I am looking into translating the Matlab to Python. There might be a few corrections that you could implement in follow-up PRs:

- It would be ideal if matlab_scripts would be numbed and named according to the architecture fig
- I notice a lot of methods under `src/dbpd/ppg`, a simple readme that describes the structure and whether the method/code is external or written by you
- `.asv` files should not be committed (add a regex to .gitignore to exclude `.asv` files) - I can fix this

PR #29, opened 2024-04-23
General refactor

kretep:
* Some renaming for clarity and general cleanup
* Moving windowing and segmenting functions to a separate module
* Factor out PreprocessingPipelineConfig


src/dbpd/windowing.py
```diff
@@ -0,0 +1,217 @@
...
+    lower_index: int
+        The dataframe index of the first sample to be windowed
+    upper_index: int
+        The dataframe index of the final sample to be windowed
+    data_point_level_cols: list
+        The columns in sensor_df that are to be kept as individual datapoints in a list instead of aggregates
+    segment_nr: int
+        The identification of the segment
+    sampling_frequency: int
+        The sampling frequency of the data
```
vedran-kasalica:
```suggestion
    sampling_frequency: int, optional
        The sampling frequency (Hz) of the data (default: 100)
```

src/dbpd/windowing.py
```diff
@@ -0,0 +1,217 @@
...
+    window_length_s: int
+        The number of seconds a window constitutes (default: 6)
+    window_step_size_s: int
+        The number of seconds between the end of the previous and the start of the next window (default: 1)
+    sampling_frequency: int
+        The sampling frequency of the data
+    segment_nr_colname: str
+        The name of the column that identifies the segment; set to None if not applicable
+    segment_nr: int
+        The identification of the segment; set to None if not applicable
```
vedran-kasalica:
```suggestion
    df: pd.DataFrame
        The original dataframe to be windowed
    time_column_name: str
        The name of the time column
    data_point_level_cols: list
        The names of the columns that are to be kept as individual datapoints in a list instead of aggregates
    window_length_s: int, optional
        The number of seconds a window constitutes (default: 6)
    window_step_size_s: int, optional
        The number of seconds between the end of the previous and the start of the next window (default: 1)
    sampling_frequency: int, optional
        The sampling frequency of the data (default: 100)
    segment_nr_colname: str, optional
        The name of the column that identifies the segment; set to None if not applicable (default: None)
    segment_nr: int, optional
        The identification of the segment; set to None if not applicable (default: None)
```

vedran-kasalica (approved):
The PR looks good, I made a few suggestions to improve the documentation. In addition, it would be useful to expose the new methods at the package level (add them to `__init__.py`). Having methods exposed on that level would make the methods easy to look up already while converting MatLab to Python.

PR #30, opened 2024-05-30
High level wrap gait

kretep:
Big refactor for gait to move all code from the notebooks into separate modules (mainly imu_preprocessing and gait_analysis). One new notebook is added that has all the processing steps as high-level function calls. In addition, several functions were extracted for repeated code where possible.

- [x] Wrap the notebooks
- [x] Fix the tests


src/dbpd/gait_analysis_config.py
```diff
@@ -0,0 +1,281 @@
...
+        
+        # TODO: generate this dictionary using object attributes (self.X) and parameters (e.g., n_dct_filters for cc)
+        self.d_channels_values: Dict[str, str] = {
+            f'grav_{self.sensor}_x_mean': 'g',
+            f'grav_{self.sensor}_y_mean': 'g',
+            f'grav_{self.sensor}_z_mean': 'g',
+            f'grav_{self.sensor}_x_std': 'g',
+            f'grav_{self.sensor}_y_std': 'g',
+            f'grav_{self.sensor}_z_std': 'g',
+            f'{self.sensor}_x_power_below_gait': 'X',
```
vedran-kasalica, line 46: We should probably specify the units, even if it's `unitless`. As it is now, a user can assume that everything labelled as `X` has the same unit, `.._z_power_below_gait`,`std_norm_acc`, etc.

src/dbpd/imu_preprocessing.py
```diff
@@ -1,10 +1,100 @@
...
+
+
+class PreprocessingConfig:
+
+    def __init__(self) -> None:
+        self.meta_filename = 'IMU_meta.json'
+        self.values_filename = 'IMU_samples.bin'
+        self.time_filename = 'IMU_time.bin'
+
+        self.acceleration_units = 'm/s^2'
```
vedran-kasalica, line 18: I think that we should add these as constants the same way we use `DataColumns.ACCELEROMETER_X`. I started it on `ppg_to_python` branch, we use `self.time_colname = DataColumns.TIME`

vedran-kasalica: I would suggest to continue it there.

docs/notebooks/gait/gait_analysis.ipynb
vedran-kasalica, line 1: I like the new structure, but I am not sure about the configuration objects as they are, we might want to include at least some parameters or have a configuration file (e.g., `toml`, or `json`) to initialize the values, as they are easy to change by the user. Especially if they just import the library and want to use the source code as it is.

Also, it would be nice to add some descriptions to the notebook. We could use markdown to describe what the cell does (maybe Eric would be able to do this more accurately). In addition, we should document the "big" methods: `preprocess_imu_data`, `extract_gait_features`, `detect_gait`, etc. They are typed, but they are missing descriptions.  I did not leave individual comments for each of the methods.

vedran-kasalica (approved):
The PR looks good, it is a big refactoring step forward. I left some comments, but they can be discussed and/or addressed in the following PRs.

docs/notebooks/gait/gait_analysis.ipynb
kretep: Not sure about the configuration files. It might also make it less clear which values are being used (the default ones, the configured ones, or the values overridden in the code...)

PR #34, opened 2024-08-14
PPG feature extract - setup the initial tests

vedran-kasalica:
The initial tests for feature extraction are added and PPG preprocessing tests were improved, however they are all still failing (as expected).


tests/test_ppg_analysis.py
```diff
@@ -14,50 +16,63 @@
...
+        output_dir_name: str = "3.extracted_features"
+        classifier_path = "src/dbpd/ppg/classifier/LR_PPG_quality.pkl"
+
+        input_path = shared_datadir / input_dir_name / "ppg"
+        reference_output_path = shared_datadir / output_dir_name / "ppg"
+
+
+        config = HeartRateFeatureExtractionConfig()
+        extract_signal_quality_features(input_path, classifier_path, reference_output_path, config)
+        compare_data(reference_output_path, reference_output_path, accelerometer_binaries_pairs)
```
kretep: Is this right? I think the function should put the output at a tested_output_path and then compare it to the reference_output_path; now it is comparing the generated data to itself

kretep (changes requested):
One possible issue with testing, looks fine otherwise

tests/test_ppg_analysis.py
vedran-kasalica: Thanks for looking into it, I updated the test!

PR #36, opened 2024-08-14
Ppg to python

kretep:
A rather large PR with part of the PPG pipeline, mainly preprocessing and feature extraction. Still a work-in-progress, but we want to start creating new topic branches with smaller scopes.


tests/data/2.preprocessed_data/PPG/IMU_new/accelerometer_meta.json
vedran-kasalica, line 1: We should probably keep one preprocessed data file, there is no need for `IMU` and `IMU_new` and the same for `PPG`

vedran-kasalica (approved):
The PR provides a big part of PPG refactoring, we still have to fix PPG preprocessing parameters. In addition, we need to finish the future extraction and classification. The tests are currently failing.

PR #37, opened 2024-08-14
Rename dbpd to paradigma

kretep:
This renames the dbpd package to paradigma. All internal references and file names should be covered. Some points of attention:
* when pulling, `poetry install` is required to re-install the package before working with it.
* a new environment will be created, which needs to be selected when running the notebooks (possibly restart your editor to find it).


docs/matlab_scripts/ppg/1-2-3_signal_quality.m
vedran-kasalica, line 1: I am not sure if these paths will work "out of the box", so Kars should look into it before re-running the Matlab code

vedran-kasalica (approved):
All the changes look good, and the tests run correctly (apart from the PPG-specific tests that were failing before the package was renamed). The PR can be merged.

PR #38, opened 2024-08-15
Update cicd workflows

kretep:
This adds the updated CI/CD workflows as we are also using for the tsdf repo. It has been tested in the paradigma-shell repo. There are still a few warnings while generating the documentation, which will be fixed later.


vedran-kasalica (approved):
Small PR. They look good, the builds are passing. I could not verify the release as it's triggered on push to the `release` branch, but I understand that it was tested earlier.

PR #41, opened 2024-08-15
Create an initial release

kretep:
The first release based on the actual code base instead of the mostly empty stub. This is also to check whether the workflows perform properly. Not really necessary to review the code, as it was already reviewed when merging into main


vedran-kasalica (approved):
The rules were already approved when merged into `main`. The features are good to be released.

PR #42, opened 2024-08-20
Expose functions at top-level

kretep:
Exposing the analysis functions for gait at the package top level. Also fixing circular imports and API documentation

### DIFF
--- a/docs/notebooks/gait/gait_analysis.ipynb
+++ b/docs/notebooks/gait/gait_analysis.ipynb
@@ -18,7 +18,10 @@
     "%autoreload 2\n",
     "\n",
     "import os\n",
-    "from paradigma import *"
+    "from paradigma.preprocessing_config import IMUPreprocessingConfig\n",
+    "from paradigma.gait_analysis import *\n",
+    "from paradigma.gait_analysis_config import *\n",
+    "from paradigma.imu_preprocessing import *"
    ]
   },
   {
--- a/docs/notebooks/gait/gait_analysis.ipynb
+++ b/docs/notebooks/gait/gait_analysis.ipynb
@@ -19,9 +19,9 @@
     "\n",
     "import os\n",
     "from paradigma.preprocessing_config import IMUPreprocessingConfig\n",
-    "from paradigma.gait_analysis import *\n",
-    "from paradigma.gait_analysis_config import *\n",
-    "from paradigma.imu_preprocessing import *"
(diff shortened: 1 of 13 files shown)


docs/notebooks/gait/gait_analysis.ipynb
Erikpostt, line 1: Why do you choose to import everything from _paradigma_? Is it not more helpful to the user if the modules used are specified during the import such that the user knows where they are defined?

src/paradigma/imu_preprocessing.py
```diff
@@ -73,8 +72,8 @@ def preprocess_imu_data(input_path: str, output_path: str, config: IMUPreprocess
 def transform_time_array(
     time_array: np.ndarray,
     scale_factor: float,
-    input_unit_type: TimeUnit,
-    output_unit_type: TimeUnit,
+    input_unit_type: paradigma.TimeUnit,
```
Erikpostt: Should the type not be _str_?

Erikpostt: This applies to multiple edited lines in this file.

Erikpostt (commented):
Hey Peter,

Looking good overall! Let me know what your view is on the comments I provided.

Best,
Erik

docs/notebooks/gait/gait_analysis.ipynb
kretep: You're right in that this may be confusing. What I tried to do here is to expose on the package level all high-level functions that are to be used by the user, but maybe we should keep them in the submodules instead. I will move part of the functions to a separate gait_util module, similar to how we are doing it for heart_rate, hopefully making a bit more clear what functions are high-level and which are helper functions. Also, I was trying to fix some cyclic imports, and this is what I ended up with, but I agree that it doesn't make it very clear what function is from where. I'll make another proposal.

src/paradigma/imu_preprocessing.py
kretep: TimeUnit is supposed to be an Enum (although I see it is not strictly one) with string values. So the type TimeUnit implies that only the values in the Enum are possible

Erikpostt: Thanks for clarifying that ;)

docs/notebooks/gait/gait_analysis.ipynb
kretep: I've reorganized the imports and moved some functions around, and now explicitly importing everything. Also tweaked the documentation configuration to only generate entries for the members that are actually defined in the module.

Erikpostt (approved):
Looking good, ready to merge from my perspective.

PR #44, opened 2024-08-29
Fix typing and improve documentation

kretep:
Creating the PR mostly to check the automated workflow


src/paradigma/gait_analysis.py
```diff
@@ -120,9 +122,9 @@ def extract_arm_swing_features(input_path: str, output_path: str, config: ArmSwi
     # perform principal component analysis on the gyroscope signals to obtain the angular velocity in the
     # direction of the swing of the arm 
     df[config.velocity_colname] = pca_transform_gyroscope(
-        df=df, 
-        y_gyro_colname=DataColumns.GYROSCOPE_Y,
-        z_gyro_colname=DataColumns.GYROSCOPE_Z,
+        df=df,
+        y_gyro_colname=DataColumns.GYROSCOPE_Y.value,
```
vedran-kasalica: I agree that using Enum is a better solution. We might want to specify the Enum types on the method level (`pca_transform_gyroscope` to expect the `DataColumns` as input). This way we don't need to use `.value` and the type checking will be enforced better.

vedran-kasalica: Looking at the other examples however I can see that setting method inputs to the `DataColumns` class would not be sufficient.

Examples such as: `arr_ppg = df_ppg[DataColumns.PPG.value].to_numpy()` would still require `.value`. 

I will look at the rest of the PR before I make any further suggestions.

src/paradigma/gait_analysis_config.py
```diff
@@ -30,10 +30,10 @@ def __init__(self) -> None:
 
         self.time_colname = 'time'
         
-        self.l_accelerometer_cols: List[str] = [
-            DataColumns.ACCELEROMETER_X, 
-            DataColumns.ACCELEROMETER_Y, 
-            DataColumns.ACCELEROMETER_Z
+        self.l_accelerometer_cols: List[str] = [ # strings, not DataColumns, as we'll edit them
```
vedran-kasalica: This part is unclear to me, why are we editing column names?

src/paradigma/gait_analysis.py
kretep: I'm actually doubting now whether the Enums are a good idea. They seemed nice for explicit typing, but they will also limit end-users from using their own custom strings...

vedran-kasalica: That's a good point. Also, the Enums don't allow much flexibility that I was hoping for. For example it is not simple to make a class `GaitColumns(DataColumn)` which extends `DataColumn(Enum)`, due to some `Enum` specific limitations.

vedran-kasalica: I tried using 

```
def __str__(self):
        return self.value
    
def __repr__(self):
        return self.value
```

to avoid the need for `.value`, however, I am getting errors when working with pd dataframes. I am getting `KeyError`. I even tried using `StrEnum` hoping that it would work better but it doesn't. I will remove `Enums`

src/paradigma/gait_analysis_config.py
vedran-kasalica: I put this under a general config (part of the config refactoring) and set them back to constants. I could not find a place where we changed them.

vedran-kasalica (approved):
I think that the PR could be merged. It cleans up the code and improves typing. The constants and config files were also improved. 
We tried using `Enum` for constants but it turned out to cause more issues without having large "typing" benefits, so we went back to using classes to group constants.

@kretep, could you look at the latest changes I made and merge them if you don't have objections?

kretep:
The additional changes look good. The enums were indeed too restrictive.

PR #53, opened 2024-09-20
Making changed based on the PD@Home paper

Erikpostt:


Erikpostt:
@kretep these updates are primarily for (1) efficiency and (2) documentation purposes. Some of these changes are cross-domain (e.g., tabulate windows), so we'll have to figure out internally later how we want to process these changes, as they may also influence the results of the other domains. Let me know what you think if you find the time.

@vedran-kasalica thank you for your suggestions earlier, the tests seem to be passing now that the test data is updated. I suggested @kretep to revier this PR since you're only working 1 day a week on this project (right?). 

And to both of you: please let me know where I can improve. Not just the code, but particularly the way I went about changing code + commiting changes + the size of the PR. At some point in time I'll have to check PRs from others internally, so it's great to hear your suggestions. ;)

Best,
Erik

kretep:
@Erikpostt it looks like some of the tests are failing; the processed data doesn't seem to match the reference anymore. I did see you modified the reference data (not sure in which commit though). Do you have an explanation? It seems it happened after you changed the CC parameters

Erikpostt:
@kretep yes this is on me. I didn't think of creating a new branch for new edits, so for now I'll close this PR and wait until the final changes are done.

kretep:
Regarding the commit style: although everybody has their own style, personally I'd say you could combine a number of commits into one. This can also be done by squashing them after you created the commits (easiest is if you didn't push yet, because that would require a force-push which is discouraged). Although it is possible to do this from the command line, a tool like '[sourcetree](https://www.sourcetreeapp.com/)' make this a bit more intuitive.

Regarding commit messages, you might take a look at some of the guides that are out there, to make them as concise and descriptive as possible: https://cbea.ms/git-commit/

kretep:
Sorry, wrong thread

PR #54, opened 2024-09-20
update test data based on change in functionalities

Erikpostt:
Hey Vedran,

What is the best way to update the test data based on changes in the pipeline? Through these PRs? It's quite annoying to have to change the code, then change the test data, merge the PR, and then push the code, isn't it? 

Cheers,
Erik


vedran-kasalica:
> Hey Vedran,
> 
> What is the best way to update the test data based on changes in the pipeline? Through these PRs? It's quite annoying to have to change the code, then change the test data, merge the PR, and then push the code, isn't it?
> 
> Cheers, Erik

Hi @Erikpostt, I am not sure I understand your question. I assume that if the pipeline is going to be changed and the changes require test data to be updated, those 2 get updated together in the same PR. Is that causing issues? 

We should just be careful, to explicitly mention in the PR that the test data is being updated (we can make a PR template with a checkboxes for that), to avoid a PR changing the test data by accident.

Erikpostt:
@vedran-kasalica that would be great indeed! Do you mean that when I change the pipeline, I should run the notebook, create new test files as a result, and add those to the PR? And then the checks will be run using this test data?

Erikpostt:
& how do you test these changes locally? If I import _paradigma_, it automatically imports the module v 0.3.2 according to Poetry. However, I want to test the updates using my local version (e.g., a the branch with the PR). How would you go about this when working with kernels?

vedran-kasalica:
> @vedran-kasalica that would be great indeed! Do you mean that when I change the pipeline, I should run the notebook, create new test files as a result, and add those to the PR? And then the checks will be run using this test data?

Yes, exactly! We should just be careful in a scenario when a change to the pipeline accidentally changes the output, and we run the notebooks, generate wrong data, and commit it accidentally. 
That is why I mentioned that stating in the PR description explicitly whether the data was changed on purpose could help the reviewer in detecting such an occurrence. 
This could also be avoided by notebooks generating new data instead of overriding the existing one. In that case, the developer would have to manually change the test data, and therefore mitigate the risk of accidental changes.

Regarding working with different versions, there might be 2 different scenarios to distinguish. If you are using `pip install paradigma` and using paradigms within a different project, that version of `paradigma` cannot be pulled from a local repo. But you can install in your local environment your local paradigma, for example by navigating to paradigma dir and calling `poetry install`.

Now, assuming that you are working on a separate branch, and run `poetry install` while it is active, the command will install the current branch to your local env. The version shown there is simply pulled from `pyproject.toml`. I would suggest changing the version on your branch (see `pyproject.toml`) to e.g., `version = "0.3.3-eric-dev"` and try to install it using `poetry install`. The message should display the new version as available. I hope this clarifies the development.

vedran-kasalica:
Is this PR still supposed to be merged or was it incorporated into #53?

Erikpostt:
@vedran-kasalica I created this PR because I thought no more changes would be necessary for the test data, but I was wrong. I'll remove this PR for now, and I'll incorporate it indeed into #53  .

PR #61, opened 2024-09-26
Don't specify datatype and bits

kretep:
These are automatically derived (and overridden) from the dataframe, so these are redundant and potentially confusing


Erikpostt (approved):
Looks good, merging.

Issue #63, opened 2024-09-30
Add functionalities for the PD@Home paper


Erikpostt:
**Problem statement**
The code of the _PD@Home_ paper shall be published together with the research paper and the dataset. To run the code smoothly, some of the functionalities of _Paradigma_ should be changed or enhanced to deal with an increasing variety of data types and formats. 

**Extra information**
The pipeline for _PD@Home_ is available at [https://github.com/biomarkersParkinson/pdathome_gait](url).

PR #64, opened 2024-09-30
Gait paper

Erikpostt:
This PR contributes features according to #63. 

@kretep this is a very large PR, and I wanted to review this code now before it got even larger. What is your take on this? I understand that normally, I would make changes for a smaller issue, which I'll try to do in the future. But what about this large contribution?

EDIT: since you have "officially" handed over the reins to us, perhaps this should be the final PR for you to review. Or, we should review this ourselves internally if you prefer.


kretep:
Code-wise this looks fine! If you'd want to improve something, maybe you could consider making your commits a bit less fine-grained. It's helpful to combine all the changes that are part of the same "theme" (in this case: at least all commits that have the same message), so that later (or when reviewing a PR) it is easier to see which changes belong together.

You could even do this after the fact by squashing the existing commits. [Here's](https://www.youtube.com/watch?v=42392W7SgnE) a helpful explanation, or otherwise a graphic tool like [Sourcetree](https://www.sourcetreeapp.com/) might help you to do that. Or you could simply wait with the commit until the changes that belong together are finished.

kretep (approved):

Erikpostt:
Thank you! I'll look into these methods for working better with commits.

PR #81, opened 2024-12-04
Imu preprocessing gyro

nienketimmermans:
Hee Erik, ik heb een aparte functie aangemaakt om alleen de gyroscope data te preprocessen. Dat vond ik voor nu het makkelijkst. Evt kan het later met de preprocess_imu_data functie gecombineerd worden.

PR #84, opened 2024-12-10
Add gait weekly aggregates

Erikpostt:
The purpose of this PR is to add (weekly) aggregates stratified by segment category to the gait pipeline as a final step. The current implementation stores a `.json` file with the arm swing parameters for different segment duration categories and overall.

Note that the arm swing quantification step in the notebook is currently commented out due to no arm swing detected in the test segment. Therefore, the suggested changes warrant validation on larger segments in the testing phase.

In the process of adding this feature, the following was modified:
- Path variable names standardized to _path_to_X_
- Notebook testing removed as it is does not fit the purpose of the release (more on this below)
- Column names standardized to `DataColumns` in _constants.py_ instead of `config`. 

### Notebook testing
The purpose of notebook testing was to ensure that the data generated in the notebooks was equal (up to a margin) to simulated data. The IO of high-level functions is currently managed inside _test_gait_analysis.py_, _test_ppg_analysis_ and _test_tremor_analysis_. Added that we should restructure the notebooks such that its goal is not to create test data, but to document and demonstrate the functionalities of Paradigma, I think we should remove the notebook testing. Furthermore note that only the _gait_ pipeline used this functionality.


Erikpostt:
@nienketimmermans if approved, don't forget to merge these changes into #81

src/paradigma/gait/gait_analysis.py
```diff
@@ -466,90 +466,106 @@ def filter_gait_io(input_path: Union[str, Path], output_path: Union[str, Path],
...
+    # Aggregate arm swing parameters per segment to obtain estimates for varying segment durations
+    df_segments = df_filtered.copy()
+    df_segments[DataColumns.SEGMENT_NR] = create_segments(
+        config=config,
+        df=df_segments
+    )
 
-#     df_segment_aggregates['segment_duration_ms'] = (df_segment_aggregates['segment_duration_s'] * 1000).round().astype(int)
-#     df_segment_aggregates = df_segment_aggregates.drop(columns=[DataColumns.SEGMENT_NR])
+    df_segments = discard_segments(
```
nienketimmermans, line 516: I would add more explanation here.

src/paradigma/gait/gait_analysis.py
```diff
@@ -466,90 +466,106 @@ def filter_gait_io(input_path: Union[str, Path], output_path: Union[str, Path],
...
+        config=config,
+        df=df_segments,
+        format='windows'
+    )
 
-#     df_weekly_median = df_filtered[['range_of_motion', f'peak_{config.velocity_colname}']].agg('median')
-#     df_weekly_quantile = df_filtered[['range_of_motion', f'peak_{config.velocity_colname}']].quantile(0.95)
-#     df_weekly_aggregates = pd.concat([df_weekly_median, df_weekly_quantile], axis=1)
+    df_segments[DataColumns.SEGMENT_CAT] = categorize_segments(df_segments, config, 'windows')
```
nienketimmermans, line 523: Here as well

nienketimmermans (changes requested):
@Erikpostt I fixed some 'path_to...' typos and added a minor suggestion to provide more explanation. After you've looked at that it's ready to be merged.

nienketimmermans (approved):

PR #85, opened 2024-12-11
Heart rate estimation

KarsVeldkamp:
Ha Erik,

Hierbij de PR voor HR estimation. Er zitten ook een paar renaming dingen in voor de preprocessing maar die spreken denk ik voor zich. Ga ik hierna mij focussen op efficiëntie ;).


docs/notebooks/ppg/ppg_analysis.ipynb
```diff
@@ -48,22 +44,31 @@
...
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Shape of the original data: (64775, 2) (72947, 4)\n",
+      "Shape of the overlapping segments: (64775, 2) (64361, 4)\n"
+     ]
+    }
+   ],
```
Erikpostt: Je hebt hier en daar nog execution counts, wat op zich niet een probleem is omdat je misschien juist deze output wilt laten zien aan een gebruiker zonder dat een gebruiker de code hoeft te draaien. Even een check of dat je intentie was.

docs/notebooks/ppg/ppg_analysis.ipynb
```diff
@@ -48,22 +44,31 @@
...
    "source": [
     "ppg_config = PPGConfig()\n",
     "imu_config = IMUConfig()\n",
     "metadatas_ppg, metadatas_imu = scan_and_sync_segments(os.path.join(path_to_sensor_data, 'ppg'),\n",
     "                                                       os.path.join(path_to_sensor_data, 'imu'))\n",
-    "df_ppg_proc, df_imu_proc=preprocess_ppg_data(metadatas_ppg[0], metadatas_imu[0],\n",
+    "df_ppg_proc, df_acc_proc=preprocess_ppg_data(metadatas_ppg[0], metadatas_imu[0],\n",
```
Erikpostt:
```suggestion
    "df_ppg_proc, df_acc_proc = preprocess_ppg_data(metadatas_ppg[0], metadatas_imu[0],\n",
```

src/paradigma/config.py
```diff
@@ -375,9 +373,8 @@ def __init__(self, min_window_length: float = 10) -> None:
         super().__init__()
 
          # Parameters for HR analysis
-        self.window_length_s: int = 6
-        self.window_step_size_s: int = 1
-        self.min_hr_samples = min_window_length * self.sampling_frequency
+        self.window_overlap_s = self.window_length_s - self.window_step_size_s
+        self.min_hr_samples = int(min_window_length * self.sampling_frequency)
```
Erikpostt: E.g., if sampling frequency equals 25 Hz, and window length equals 3.5, then we have  `min_window_length * self.sampling_frequency` equals 25 * 3.5 = 87.5. Now, inconveniently `int(87.5)` equals 87 (`int` rounds to the lowest integer). So perhaps, for these edge cases, you want to use `int(np.round(min_window_length * self.sampling_frequency))` or something along these lines.

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -85,4 +87,60 @@ def signal_quality_classification_io(input_path: Union[str, Path], output_path:
     metadata_time, metadata_values = read_metadata(input_path, config.meta_filename, config.time_filename, config.values_filename)
     df = tsdf.load_dataframe_from_binaries([metadata_time, metadata_values], tsdf.constants.ConcatenationType.columns)
 
-    df = signal_quality_classification(df, config, path_to_classifier_input)
\ No newline at end of file
+    df = signal_quality_classification(df, config, path_to_classifier_input)
+
+
+def estimate_heart_rate(df: pd.DataFrame, df_ppg_preprocessed: pd.DataFrame, config:HeartRateExtractionConfig) -> pd.DataFrame:
```
Erikpostt: To me it's a little ambiguous as to the meaning of `df`. Could you perhaps choose a different naming for `df`, and expand on the difference between these two dataframes in the docstrings?

Erikpostt:
```suggestion
def estimate_heart_rate(df: pd.DataFrame, df_ppg_preprocessed: pd.DataFrame, config: HeartRateExtractionConfig) -> pd.DataFrame:  
```

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -85,4 +87,60 @@ def signal_quality_classification_io(input_path: Union[str, Path], output_path:
...
+
+    Returns
+    -------
+    pd.DataFrame
+        The DataFrame containing the heart rate estimations.
+    """
+       
+    # Assign window-level probabilities to individual samples
+    ppg_post_prob = df.loc[:, DataColumns.PRED_SQA_PROBA].to_numpy()
+    #acc_label = df.loc[:, DataColumns.ACCELEROMETER_LABEL].to_numpy() # Adjust later in data columns to get the correct label, should be first intergrated in feature extraction and classification
```
Erikpostt: Is this intentional?

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -85,4 +87,60 @@ def signal_quality_classification_io(input_path: Union[str, Path], output_path:
...
+
+    sqa_label = assign_sqa_label(ppg_post_prob, config)
+    v_start_idx, v_end_idx = extract_hr_segments(sqa_label, config.min_hr_samples)
```
Erikpostt: Perhaps add brief comments (#) here to explain what happens, e.g., `# Add signal quality value to each window` and `# Extract HR segments above signal quality threshold`.

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -85,4 +87,60 @@ def signal_quality_classification_io(input_path: Union[str, Path], output_path:
...
+    
+    fs = config.sampling_frequency
+
+    v_hr_rel = np.array([])
+    t_hr_rel = np.array([])
+
+    for start_idx, end_idx in zip(v_start_idx, v_end_idx):
+        # Skip if the epoch cannot be extended by 2s on both sides
+        if start_idx < 2 * fs or end_idx > len(df_ppg_preprocessed) - 2 * fs:
```
Erikpostt: You seem to be using `2 * fs` quite often, perhaps you can assign this a variable once.

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -85,4 +87,60 @@ def signal_quality_classification_io(input_path: Union[str, Path], output_path:
...
+            continue
+        
+        # Extract the extended PPG segment for HR estimation
+        extended_ppg_segment = df_ppg_preprocessed[DataColumns.PPG][start_idx - 2 * fs : end_idx + 2 * fs]
+
+        # Perform HR estimation
+        hr_est = extract_hr_from_segment(extended_ppg_segment, config.tfd_length, fs, config.kern_type, config.kern_params)
+
+        # Generate HR estimation time array
+        rel_segment_time = df_ppg_preprocessed.time[start_idx:end_idx].values
```
Erikpostt: Is `.time` the same as `[DataColumns.TIME]`?

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -85,4 +87,60 @@ def signal_quality_classification_io(input_path: Union[str, Path], output_path:
...
+        n_full_segments = len(rel_segment_time) // config.hr_est_samples
+        hr_time = rel_segment_time[:n_full_segments * config.hr_est_samples : config.hr_est_samples] # relative time in seconds after the start of the segment
```
Erikpostt: Perhaps you could clarify this process a little using comments (#), I'm not sure I understand it as it is now.

src/paradigma/heart_rate/heart_rate_analysis_config.py
Erikpostt, line 1: Why did you create this config again? Is it necessary for your code to run? Or is this your way of showing you disagree with the current format? ;)

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
+import numpy as np
+from typing import Tuple
+
+from paradigma.config import HeartRateExtractionConfig
+from paradigma.heart_rate.tfd import nonsep_gdtfd
+
+def assign_sqa_label(ppg_prob, config: HeartRateExtractionConfig, acc_label=None) -> np.ndarray:
```
Erikpostt:
```suggestion
def assign_sqa_label(ppg_prob: np.ndarray, config: HeartRateExtractionConfig, acc_label: np.ndarray=None) -> np.ndarray:
```

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+    Assigns a signal quality label to every individual data point.
+
+    Parameters:
+    - ppg_prob: numpy array of probabilities for PPG
+    - acc_label: numpy array of labels for accelerometer (optional)
+    - fs: Sampling frequency
```
Erikpostt: I think these need to be updated.

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+        acc_label = np.ones(len(ppg_prob))
+
+
+    # Number of samples in an epoch
+    fs = config.sampling_frequency
+    samples_per_epoch = config.window_length_s * fs
+
+    # Calculate number of samples to shift for each epoch
+    samples_shift = config.window_step_size_s * fs
+    n_samples = int((len(ppg_prob) + config.window_overlap_s) * fs)
```
Erikpostt: Same to a comment I made earlier, perhaps you first want to round before converting to an integer to avoid 8.6 being rounded to 8.

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+    data_prob = np.zeros(n_samples)
+    data_label_imu = np.zeros(n_samples, dtype=np.int8)
+
+    for i in range(n_samples):
+        # Start and end indices for current epoch
+        start_idx = int(np.floor((i - (samples_per_epoch - samples_shift)) / fs))
+        end_idx = int(np.floor(i / fs))
```
Erikpostt: I believe `int` already does `np.floor` inherently, but you could check.

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Correct for first and last epochs
+        if start_idx < 0:
+            start_idx = 0
```
Erikpostt: Perhaps, instead of this, you could change the earlier:

`start_idx = int(np.floor((i - (samples_per_epoch - samples_shift)) / fs))`

to:

`start_idx = max(0, int((i - (samples_per_epoch - samples_shift)) / fs))`

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+        if end_idx > len(ppg_prob):
+            end_idx = len(ppg_prob)
```
Erikpostt: Similar approach could be used here.

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # Extract probabilities and labels for the current epoch
+        prob = ppg_prob[start_idx:end_idx+1]
+        label_imu = acc_label[start_idx:end_idx+1]
+
+        # Calculate mean probability and majority voting for labels
+        data_prob[i] = np.mean(prob)
+        data_label_imu[i] = int(np.mean(label_imu) >= 0.5)
```
Erikpostt, line 45: Use round here as well if appropriate.

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+
+    # Set probability to zero if majority IMU label is 0
+    data_prob[data_label_imu == 0] = 0
+    sqa_label = (data_prob > config.threshold_sqa).astype(np.int8)
```
Erikpostt: Two questions:
1. Should the probability be (1) larger than, or (2) larger than or equal to?
2. Why do you convert it to `np.int8`? Why don't you store it as a boolean?

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Parameters
+    ----------
+    ppg : np.ndarray
+        The preprocessed PPG segment with 2 seconds of padding on both sides to reduce boundary effects.
+    tfd_length : int
+        Length of each segment (in seconds) to calculate the time-frequency distribution.
+    fs : int
+        The sampling frequency of the PPG signal.
+    MA : int
```
Erikpostt: I think you don't use this one anymore.

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+        kernels like 'spwvd' or 'swvd'. Default is None.
+
+    Returns
+    -------
+    np.ndarray
+        The estimated heart rate.
+    """
+
+    # Constants to handle boundary effects
+    edge_padding = 4  # Additional 4 seconds (2 seconds on both sides)
```
Erikpostt: I'd suggest assigning `edge_padding = 4 * fs` and `tfd_length = tfd_length * fs`, since you use `edge_padding` and `tfd_length` only in combination with `fs` in this function, right?

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+    extended_epoch_length = tfd_length + edge_padding
+
+    # Calculate the actual segment length without padding
+    original_segment_length = (len(ppg) - edge_padding * fs) / fs
+
+    # Determine the number of tfd_length-second segments
+    if original_segment_length > extended_epoch_length:
```
Erikpostt: Larger, or larger or equal to?

src/paradigma/heart_rate/heart_rate_estimation.py
```diff
@@ -0,0 +1,185 @@
...
+    fs : int
+        Sampling frequency in Hz.
+    kern_type : str
+        Type of TFD kernel to use.
+    kern_params : dict
+        Parameters for the specified kernel.
+
+    Returns
+    -------
+    hr_smooth_tfd : list
```
Erikpostt: `np.array` or `np.ndarray` now I believe.

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+    - get_window: General function to calculate a window function.
+    - get_win: Helper function to create the specified window type.
+    - shift_window: Shifts the window so that positive indices appear first.
+    - pad_window: Zero-pads the window to a specified length.
+    - compute_tfd: Finalizes the time-frequency distribution computation.
+
+
+"""	
+
+def nonsep_gdtfd(x, kern_type=None, kern_params=None):
```
Erikpostt: We have used `Union` so far, but apparently the correct way since Python 3.10 is using Pipe: 

```suggestion
def nonsep_gdtfd(
x: np.ndarray, 
kern_type: None | str = None, 
kern_params: None | dict = None
):
```

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+        Analytic signal in the time domain with zeroed second half.
+    """
+    N = len(x)
+    
+    # Zero-pad the signal to double its length
+    x = np.concatenate((np.real(x), np.zeros(N)))
+    x_fft = np.fft.fft(x)
+
+    # Generate the analytic signal in the frequency domain
+    H = [1] + list(np.repeat(2, N-1)) + [1] + list(np.repeat(0, N-1))
```
Erikpostt: Perhaps you could use `np.concatenate` here, which works for more than 2 items as well: 

```
# Define three arrays
arr1 = np.array([1, 2, 3])
arr2 = np.array([4, 5, 6])
arr3 = np.array([7, 8, 9])

# Concatenate along the first axis (default)
result = np.concatenate((arr1, arr2, arr3))
print(result)  # Output: [1 2 3 4 5 6 7 8 9]
```

Or for lists:

```
# Define three lists
list1 = [1, 2, 3]
list2 = [4, 5, 6]
list3 = [7, 8, 9]

# Concatenate lists with NumPy
result = np.concatenate((list1, list2, list3))
print(result)  # Output: [1 2 3 4 5 6 7 8 9]
```

I suppose that the FFT works faster with two numpy arrays, but not sure.

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+    z_cb = np.fft.ifft(x_fft * H)
+
+    # Force the second half of the time-domain signal to zero
+    z = np.concatenate((z_cb[:N], np.zeros(N)))
+
+    return z
+
+def gen_time_lag(z):
```
Erikpostt:
```suggestion
def gen_time_lag(z: np.ndarray): -> np.ndarray
```

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+        # Extract the time slice from the analytic signal
+        K_time_slice = z[inp] * np.conj(z[inn])
+
+        # Store real and imaginary parts
+        tfd[n, :Nh] = np.real(K_time_slice)
+        tfd[n, Nh:] = np.imag(K_time_slice)
+    
+    return tfd
+
+def multiply_kernel_signal(tfd, kern_type, kern_params, N, Nh):
```
Erikpostt:
```suggestion
def multiply_kernel_signal(
tfd: np.ndarray, 
kern_type: str, 
kern_params: dict, 
N: int, 
Nh: int
) -> np.ndarray:
```

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+    tfd : ndarray
+        Modified TFD after kernel multiplication.
+    """
+    # Loop over lag indices
+    for m in range(Nh):
+        # Generate the Doppler-lag kernel for each lag index
+        g_lag_slice = gen_doppler_lag_kern(kern_type, kern_params, N, m)
+        
+        # Extract and transform the TFD slice for this lag
+        tfd_slice = np.fft.fft(tfd[:, m]) + 1j * np.fft.fft(tfd[:, Nh + m])
```
Erikpostt, line 259:
```suggestion
        tfd_slice = np.fft.fft(tfd[:, m]) + j * np.fft.fft(tfd[:, Nh + m])
```

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+        # Multiply by the kernel and perform inverse FFT
+        R_lag_slice = np.fft.ifft(tfd_slice * g_lag_slice)
+        
+        # Store real and imaginary parts back into the TFD
+        tfd[:, m] = np.real(R_lag_slice)
+        tfd[:, Nh + m] = np.imag(R_lag_slice)
+    
+    return tfd
+
+def gen_doppler_lag_kern(kern_type, kern_params, N, lag_index):
```
Erikpostt: Also fix parameter types in your other functions.

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+        win_length = kern_params['win_length']
+        win_type = kern_params['win_type']
+        win_param = kern_params['win_param'] if l >= 3 else 0
+        win_param2 = kern_params['win_param2'] if l >= 4 else 0
+
+        G2 = get_window(win_length, win_type, win_param) # Generate the window, same per lag iteration
+        G2 = pad_window(G2, N)      # Zero-pad the window to the length of the signal
+        G2 = G2[lag_index]          # Extract the lag_index-th element of the window --> can this be done more efficiently? Since we only need one element of the window and the padding is similar for every iteration
+
+        g[:] = G2 # Assign the window to the kernel
```
Erikpostt: I think you could combine these:

First:
```
if kern_type not in [X, Y, Z, A, B, C]:
    raise("You should pick one of ...")
```

Note: you can also use it at the end like you've used it in the `else` statement. I only saw this later. 

Then
```
if kern_type == 'wvd':
    ...
```

```
elif kern_type == 'sep': 
    ....
```

```
else:
    if l < 2:
        raise ValueError("Need at least two window parameters for {kern_type.upper()}")

    if kern_type == 'swvd':
        key = 'lag'
    else:
        key = 'doppler'
        
      win_length = kern_params['win_length']
      win_type = kern_params['win_type']
      win_param = kern_params['win_param'] if l >= 3 else 0
      win_param2 = kern_params['win_param2'] if l >= 4 else 1

      G = get_window(win_length, win_type, win_param)
      G = pad_window(G, N)

      if kern_type == 'sqvd' and win_param2 == 0:
          G = np.fft.fft(G)
          G /= G[0]
    elif kern_type == 'pwvd':
        G = G[lag_index]    

      g[:] = G 
```

And I don't think you have to assign `key` in this loop because you don't seem to be using it.

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+        If greater than 0, zero-pads the window to length Npad. Default is 0.
+    
+    Returns:
+    --------
+    win : ndarray
+        The calculated window (or its DFT if dft_window is True).
+    """
+    # Handle optional arguments
+    if win_param is None:
+        win_param = []
```
Erikpostt: Perhaps you could set the parameter `win_param: list = []`.

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+    
+    # Get the window
+    win = get_win(win_length, win_type, win_param, dft_window)
```
Erikpostt, line 387: I see your `get_win` function here assigns an empty list to `win_param` if it equals `None`. However, in the `get_win()` function you assign floats to this variable. Is this working as expected?

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+    """
+    m = np.arange(0, Nh)  # m = 0:(Nh-1)
+    mb = np.arange(1, Nh)  # mb = 1:(Nh-1)
+
+    m_real = np.arange(Nh) # m_real = 0:(Nh-1) --> is this the same as m? In matlab it is different because of 1-based indexing
+    m_imag = np.arange(Nh, N) # m_imag = Nh:(N-1)
+
+    for n in range(0, N-1, 2):  # for n=0:2:N-2
+        R_even_half = np.complex128(tfd[n, :Nh]) + 1j * np.complex128(tfd[n, Nh:])
+        R_odd_half = np.complex128(tfd[n+1, :Nh]) + 1j * np.complex128(tfd[n+1, Nh:])
```
Erikpostt, line 530:
```suggestion
        R_even_half = np.complex128(tfd[n, :Nh]) + j * np.complex128(tfd[n, Nh:])
        R_odd_half = np.complex128(tfd[n+1, :Nh]) + j * np.complex128(tfd[n+1, Nh:])
```

src/paradigma/heart_rate/tfd.py
```diff
@@ -0,0 +1,553 @@
...
+        R_tslice_odd = np.zeros(N, dtype=np.complex128)
+
+        R_tslice_even[m] = R_even_half
+        R_tslice_odd[m] = R_odd_half
+
+        R_tslice_even[N-mb] = np.conj(R_even_half[mb])  
+        R_tslice_odd[N-mb] = np.conj(R_odd_half[mb])
+        
+        # Perform FFT to compute time slices
+        tfd_time_slice = np.fft.fft(R_tslice_even + 1j * R_tslice_odd)
```
Erikpostt, line 542:
```suggestion
        tfd_time_slice = np.fft.fft(R_tslice_even + j * R_tslice_odd)
```

src/paradigma/ppg_preprocessing.py
```diff
@@ -288,23 +289,23 @@ def synchronization(ppg_meta, imu_meta):
 
     return segment_ppg_total, segment_imu_total
 
-def extract_overlapping_segments(df_ppg, df_imu, time_column_ppg='time', time_column_imu='time'):
+def extract_overlapping_segments(df_ppg, df_acc, time_column_ppg='time', time_column_imu='time'):
```
Erikpostt, line 292: Don't forget to set data types here too.

Erikpostt (changes requested):
Looking good! I have a couple of suggestions, mostly related to fixing data types and simplifying code. I would also suggest to add more comments (#) in the code to make sure the user/dev understands the processing of the data.

docs/notebooks/ppg/ppg_analysis.ipynb
KarsVeldkamp: Ehm uiteindelijk wel, maar niet voor nu eig dus excuses

src/paradigma/config.py
KarsVeldkamp: Goed punt! En zal het aanpassen, heb je bezwaren tegen round i.p.v. np.round omdat je anders in de config numpy moet inladen als module @Erikpostt

src/paradigma/heart_rate/heart_rate_analysis.py
KarsVeldkamp: bedoel je dat de acc_label uitgecomment is? Want dat is intentional nu ja @Erikpostt

src/paradigma/heart_rate/heart_rate_analysis.py
KarsVeldkamp: nee, goed punt

src/paradigma/heart_rate/heart_rate_analysis_config.py
KarsVeldkamp: Om jou te pesten, nee zonder gekheid ik had hem nog nodig om de config.py aan te passen omdat hij in de vorige merge was verwijderd. Echter had ik hem al gecommit naar deze branch voordat we de configs gingen omschrijven dus daardoor is hij uit het oog verloren. Zal hem verwijderen @Erikpostt

src/paradigma/config.py
Erikpostt: There are some [minor differences](https://stackoverflow.com/questions/20388071/what-are-the-under-the-hood-differences-between-round-and-numpy-round) between the two methods, and I don't think that having to import `numpy` should restrict us to using `round()`. I'll let you decide on this matter.

src/paradigma/heart_rate/heart_rate_analysis.py
Erikpostt: Yes, I was just wondering.

src/paradigma/config.py
KarsVeldkamp: ALlright, but it doesn't seem that np.round is preferred over round() right? So than we can use just round() @Erikpostt

Erikpostt: No they seem to have different methods for rounding, but I expect that, in case the rounding is ever done differently, it should not affect the results much.

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: Not for negative numbers, and that's what I need, so I will update it to integer division because that's probably the most efficient

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: In this case not, since it is a true false statement which I want to convert to 0 or 1, therefore this doesn't have to be rounded

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: 1. Probability is larger than (1) instead of (2) in my Matlab code, however this is not entirely correct so I would want to have (2) so I adjust it here. I won't expect that this really matters for the results because I won't expect that many probabilities to be exactly the threshold
2. I'll update this and store it as a boolean

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: correct

src/paradigma/heart_rate/heart_rate_estimation.py
Erikpostt: Could `i - (samples_per_epoch - samples_shift)` be negative? Smart thinking though about the `np.floor` for negative values.

src/paradigma/heart_rate/heart_rate_estimation.py
Erikpostt: Haha you're right, not sure where this suggestion came from. You might want to store it as `boolean` if TSDF allows (don't think it does yet though ...)

src/paradigma/heart_rate/heart_rate_estimation.py
Erikpostt: Great! You might run into issues with storing as boolean due to TSDF constraints. I'm sorry if I got your hopes up.

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: Yes it is for the edge cases at the start where i = 1:150 (because samples_per_epoch is 180 and samples_shift is 30)

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: This step doesn't have to be stored in tsdf luckily :). It's just an input for extracting the ppg segments suited for HR estimation

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: agree

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: changed it to max function

src/paradigma/heart_rate/tfd.py
KarsVeldkamp: Agree, i did rewrite the function accordingly

src/paradigma/heart_rate/tfd.py
KarsVeldkamp: I think I can just remove this, since i don't have to enforce win_param to be an empty list since the get_win function handles a none value correctly

src/paradigma/heart_rate/tfd.py
KarsVeldkamp: see previous comment

src/paradigma/heart_rate/tfd.py
KarsVeldkamp: 1j is complex unit while j causes an error since it is not default in python

src/paradigma/heart_rate/tfd.py
KarsVeldkamp: I want 1j since it is a complex unit, while j is not by default

src/paradigma/heart_rate/heart_rate_estimation.py
KarsVeldkamp: I have ajdusted the ppg_prob suggestion but not for the acc_label since I got a type error from this assignment

src/paradigma/ppg_preprocessing.py
KarsVeldkamp: This is not up to date anymore I guess since we are combining stuff in the other PR?

KarsVeldkamp:
I finally made the relevant adjustments so its up to you again @Erikpostt

docs/notebooks/ppg/ppg_analysis.ipynb
```diff
@@ -9,23 +9,19 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
```
Erikpostt: Execution counts are still visible, always difficult ;)

docs/notebooks/ppg/ppg_analysis.ipynb
```diff
@@ -56,14 +52,14 @@
     "imu_config = IMUConfig()\n",
     "metadatas_ppg, metadatas_imu = scan_and_sync_segments(os.path.join(path_to_sensor_data, 'ppg'),\n",
     "                                                       os.path.join(path_to_sensor_data, 'imu'))\n",
-    "df_ppg_proc, df_imu_proc=preprocess_ppg_data(metadatas_ppg[0], metadatas_imu[0],\n",
+    "df_ppg_proc, df_acc_proc = preprocess_ppg_data(metadatas_ppg[0], metadatas_imu[0],\n",
```
Erikpostt: The indentation is a little off in this cell block. It's difficult to provide suggestions, but in general for longer lines of code using functions, try entering a new line after the opening bracket "(" and a final line for the closing bracket ")".

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -64,25 +65,84 @@ def signal_quality_classification(df: pd.DataFrame, config: SignalQualityClassif
 
     # Prepare the data
     lr_clf.feature_names_in_ = ['var', 'mean', 'median', 'kurtosis', 'skewness', 'f_dom', 'rel_power', 'spectral_entropy', 'signal_to_noise', 'auto_corr']
-    X = df.loc[:, lr_clf.feature_names_in_]
+    X = df_windowed.loc[:, lr_clf.feature_names_in_]
 
     # Normalize features using mu and sigma
     X_normalized = X.copy()
     for idx, feature in enumerate(lr_clf.feature_names_in_):
         X_normalized[feature] = (X[feature] - mu[idx]) / sigma[idx]
```
Erikpostt: I think this the can be serialized using `numpy` or `scipy`, but perhaps for another point in time. Possibly the same holds for the ratio.

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -64,25 +65,84 @@ def signal_quality_classification(df: pd.DataFrame, config: SignalQualityClassif
 
     # Prepare the data
     lr_clf.feature_names_in_ = ['var', 'mean', 'median', 'kurtosis', 'skewness', 'f_dom', 'rel_power', 'spectral_entropy', 'signal_to_noise', 'auto_corr']
```
Erikpostt, line 67: I had a discussion before about normalizing the dominant frequency, and we decided to not scale it because the frequencies can simply be different for different people. Not sure how this works for you.

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -64,25 +65,84 @@ def signal_quality_classification(df: pd.DataFrame, config: SignalQualityClassif
...
         X_normalized[feature] = (X[feature] - mu[idx]) / sigma[idx]
 
     # Make predictions for PPG signal quality assessment
-    df[DataColumns.PRED_SQA_PROBA] = lr_clf.predict_proba(X_normalized)[:, 0]                   
-    df.drop(columns = lr_clf.feature_names_in_, inplace=True)  # Drop the features used for classification since they are no longer needed
-    
-    return df    
+    df_sqa = df_windowed.copy()
+    df_sqa[DataColumns.PRED_SQA_PROBA] = lr_clf.predict_proba(X_normalized)[:, 0]                   
+    df_sqa.drop(columns = lr_clf.feature_names_in_, inplace=True)  # Drop the features used for classification since they are no longer needed
```
Erikpostt: New best practices omit using `inplace` and instead recommend using direct object assignment. Again, not necessary per se, just a recommendation.

src/paradigma/heart_rate/heart_rate_analysis.py
```diff
@@ -64,25 +65,84 @@ def signal_quality_classification(df: pd.DataFrame, config: SignalQualityClassif
...
+        # Perform HR estimation
+        hr_est = extract_hr_from_segment(extended_ppg_segment, config.tfd_length, fs, config.kern_type, config.kern_params)
+
+        # Generate HR estimation time array
+        rel_segment_time = df_ppg_preprocessed[DataColumns.TIME][start_idx:end_idx].values # relative time in seconds after the start of the segment of each sample
+        n_full_segments = len(rel_segment_time) // config.hr_est_samples # number of full segments
+        hr_time = rel_segment_time[:n_full_segments * config.hr_est_samples : config.hr_est_samples] # relative time extracted from the full segments of every 2 seconds HR estimation
+
+        # Concatenate HR estimations and times
+        v_hr_rel = np.concatenate((v_hr_rel, hr_est))
```
Erikpostt: Concatenation of arrays can be inefficient when working with many loops. There is a smarter way by creating an empty array using `np.empty()` and setting the size beforehand in the empty call. Then, in the loop, you can add items to specific indices.

src/paradigma/heart_rate/tfd.py
```diff
@@ -170,15 +173,15 @@ def gen_analytic(x):
     x_fft = np.fft.fft(x)
 
     # Generate the analytic signal in the frequency domain
-    H = [1] + list(np.repeat(2, N-1)) + [1] + list(np.repeat(0, N-1))
+    H = np.concatenate(([1], np.repeat(2, N-1), [1], np.repeat(0, N-1)))
```
Erikpostt: You're not going to like this, but I've browsed online and the usage of `np.repeat` is not so efficient, so perhaps better to go back to the old format:

H = np.empty(2 * N)  # Preallocate an array of size 2*N
H[0] = 1            # First element
H[1:N] = 2          # Next N-1 elements
H[N] = 1            # Middle element
H[N+1:] = 0         # Last N-1 elements

But I'll leave that to you for now.

Erikpostt (approved):
Looking good. I've made some suggestions, but they should not block the PR. Feel free to adjust where you agree, and otherwise merge the changes.

docs/notebooks/ppg/ppg_analysis.ipynb
KarsVeldkamp: I don't know what happened here, since some cells are null while others are not... But now I updated this

docs/notebooks/ppg/ppg_analysis.ipynb
KarsVeldkamp: Agree, I have created now a single line again since it is also not that long

src/paradigma/heart_rate/heart_rate_analysis.py
KarsVeldkamp: Apologies, you made this suggestion before and I promised to fix this in this PR but totally forgot. Will update this ;)

src/paradigma/heart_rate/heart_rate_analysis.py
KarsVeldkamp: I trained my classifier on a normalized dominant frequency for now, but I see the point. Lets discuss this another time ;)

src/paradigma/heart_rate/heart_rate_analysis.py
KarsVeldkamp: Agree, I updated this and chained the operation of copying, assigning and dropping in one to keep it concise, clear and functional

src/paradigma/heart_rate/tfd.py
KarsVeldkamp: Haha, I changed it back, since I experimented a bit and the allocated suggestion is twice as fast

KarsVeldkamp:
I have made the final adjustments and going to merge now

PR #97, opened 2025-01-08
Documentation tremor pipeline

nienketimmermans:
@Erikpostt I added the tutorial for the tremor pipeline. I also modified the mfcc calculation, so this should work the same as in the gait pipeline now. For now it's still a separate function, as your changes were not merged yet.


docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,968 @@
...
+   "metadata": {},
+   "source": [
+    "# Tremor analysis"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial shows how to run the tremor pipeline on prepared IMU or gyroscope data to obtain weekly aggregated tremor measures."
```
Erikpostt: Are the aggregations at week-level, or simply at the level of the provided dataframe?

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,968 @@
...
+   "metadata": {},
+   "source": [
+    "## Load example data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Example IMU data (8 minutes) from a participant of the Personalized Parkinson Project is loaded. The data was prepared as explained in `data_preparation.ipynb`. The prepared data contains both accelerometer and gyroscope data, but only gyroscope data is necessary for running the tremor pipeline."
```
Erikpostt, line 28: It might be worth stating that imu sensors collect data at a fixed sampling frequency, but the sampling rate is not uniform, causing variation in time differences between timestamps (which can clearly be observed in the dataframe). The `resample_data` function therefore resamples the timestamps to be uniformly distributed, and then interpolates gyroscope values at these new timestamps using the original timestamps and corresponding gyroscope values. Something along these lines perhaps?

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,968 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 2: Extract tremor features"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The preprocessed gyroscope data is windowed (window length is specified in `config`) and for each window the tremor feautures are extracted. The features consist of 12 mel-frequency cepstral coefficients (MFCCs), the frequency of the peak in the estimated power spectral density, the low frequency power (0.5 - 3 Hz power) and the tremor power."
```
Erikpostt: I'd rephrase it a little, but feel free to phrase it to your preference:

> The preprocessed gyroscope data is windowed using non-overlapping windows of length `config.window_length_s` such that features can be extracted. Features extracted from these windows are 12 mel-frequency cepstral coefficients (MFCCs), frequency of the peak in the power spectral density, power in lower frequencies (0.5 -- 3 Hz), and power in the tremor band (X -- Y Hz).

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Oh yes weekly should be removed here.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,968 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 3: Detect tremor"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A pretrained logistic regression classifier is used to predict the tremor probability (`pred_tremor_proba`) for each window, based on the MFCCs. Using the prespecified threshold, a tremor label of 0 (no tremor) or 1 (tremor) is assigned (`pred_tremor_logreg`). Next, two extra criteria for rest tremor are applied: the frequency of the peak should be between 3-7 Hz and the low frequency power should be below a prespecified threshold. The final tremor label is saved in `pred_tremor_checked`."
```
Erikpostt: - As a user, I'd be confused why more features were extracted than used in this step. Could you clarify?
- Why is there a prespecified threshold?
- Perhaps also print the threshold

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,968 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 4: Compute aggregated tremor measures"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The final step is to compute the amount of tremor time and tremor power, aggregated over all windows in the input dataframe.\n",
```
Erikpostt, line 944: - I think a user might be interested in what the tremor power is. What does this say about the severity of tremor? 

> "Tremor time is calculated as the number of the detected tremor windows, as percentage of the number of windows \n",
    "while the arm is at rest or in stable posture (when the low frequency power is below a specified threshold).

- I don't quite understand this. So you first subset all windows based on the low frequency power, and then you take the percentage of these windows in which tremor was detected?

> For tremor power the following aggregates are derived: the median, mode and 90th percentile of tremor power. The aggregated tremor measures and metadata are stored in a json file.

- As a user, I might want to be able to specify which aggregates I want to extract. Perhaps you could add a parameter for this somewhere? And then show the output for the aggregates you propose?

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,968 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
```
Erikpostt: You might want to consider [pprint](https://docs.python.org/3.8/library/pprint.html) for this.

src/paradigma/tremor/tremor_analysis.py
```diff
@@ -196,8 +196,8 @@ def detect_tremor_io(input_path: Union[str, Path], output_path: Union[str, Path]
 def aggregate_tremor_power(tremor_power: pd.Series, config: TremorQuantificationConfig):
```
Erikpostt: I raised this before in your tutorial, but I think this is the place you might want to add some flexibility to which parameters are extracted.

src/paradigma/tremor/tremor_analysis.py
```diff
@@ -196,8 +196,8 @@ def detect_tremor_io(input_path: Union[str, Path], output_path: Union[str, Path]
...
     
     """
-    Converts the tremor power to the log scale and subsequentyly computes three aggregates of tremor power:
-    the median, mode and perdentile specified in TremorQuantificationConfig.     
+    Converts the tremor power to the log scale and subsequently computes three aggregates of tremor power:
+    the median, mode and percentile specified in TremorQuantificationConfig.     
     """
 
     tremor_power = np.log10(tremor_power+1) # convert to log scale
```
Erikpostt: For consistency, and I guess also for accuracy, you might want to add `1e-10` here instead of 1.

src/paradigma/tremor/tremor_analysis.py
```diff
@@ -264,7 +264,7 @@ def quantify_tremor(df: pd.DataFrame, config: TremorQuantificationConfig):
             'nr_windows_total': nr_windows_total,
             'nr_windows_rest': nr_windows_rest
         },
-        'weekly_tremor_measures': {
+        'aggregated_tremor_measures': {
             'tremor_time': tremor_time,
```
Erikpostt: Perhaps you could make `'tremor_time'` a little more self-explanatory, such as: `'perc_windows_tremor'`, or `'tremor_time_s'`.

Erikpostt (changes requested):
Nice tutorial, I think this is a great step forward. I've made some suggestions to your changes, let me know what you think.

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Do you mean that the tremor power is not used here? 
The threshold was determined when training the algorithm, so it's similar to the threshold for the logistic regression algorithm. We can refer to the paper as soon as it's published, but that might take a while. Do you think it needs more explanation here?

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: > I don't quite understand this. So you first subset all windows based on the low frequency power, and then you take the percentage of these windows in which tremor was detected?

Yes that's right, if a person is moving their arm they can't have rest tremor at that moment, so it's a way to control for the amount of arm movement during a day or week. 

> As a user, I might want to be able to specify which aggregates I want to extract. Perhaps you could add a parameter for this somewhere? And then show the output for the aggregates you propose?

Yes the 90th percentile is actually a parameter that could be changed, so I will make that clear here.

src/paradigma/tremor/tremor_analysis.py
nienketimmermans: The 90th percentile is flexible. I can also make the median flexible, but I'm not sure whether that is useful. Do you think it's a good idea to add more options? Or what do you mean exactly by adding more flexibility?

src/paradigma/tremor/tremor_analysis.py
nienketimmermans: The 1 is added here to make the output more convenient: a value of 0 in log scale then corresponds to 0 tremor power in the original scale.

src/paradigma/tremor/tremor_analysis.py
nienketimmermans: Good idea

nienketimmermans:
@Erikpostt Thanks for the comments, I just posted a few questions to clarify some things

src/paradigma/tremor/tremor_analysis.py
nienketimmermans: Maybe I can use your `aggregate_parameter` function and integrate the mode there as well?

docs/tutorials/tremor_analysis.ipynb
Erikpostt: - In the cell before this, you state that you extract 12 MFCCs, frequency of the peak in the power spectral density, power in lower frequencies (0.5 -- 3 Hz), and power in the tremor band. In this cell, you state that you only use MFCCs for predicting tremor probability. That's why a user may be confused here.
- I personally understand where the threshold came from, but a new user may not. You could perhaps state that the threshold was set to minimize false-positives when training the classifier? Or something along these lines? By printing the threshold, a user can decide whether to change the threshold if desired. In the future we can refer to the paper indeed. :)

docs/tutorials/tremor_analysis.ipynb
Erikpostt: > Yes that's right, if a person is moving their arm they can't have rest tremor at that moment, so it's a way to control for the amount of arm movement during a day or week.

I see, thanks for clarifying. You might want to expand slightly on this so that it is clear to a user. The line was a little confusing to me.

src/paradigma/tremor/tremor_analysis.py
Erikpostt: Yes I think we could ensure the `aggregate_parameter` works for all domains. That way, users can decide themselves which parameters to extract. That is also what I mean by "adding more flexibility". Feel free to add the mode and perhaps define it in a central location, such as `util.py`.

src/paradigma/tremor/tremor_analysis.py
Erikpostt: Ah I see, that makes sense. I'll leave it to you, but you might want to clarify this decision here.

nienketimmermans: I mentioned it in the `aggregate_tremor` function, but that's indeed not very close to the line of code. As I will use your `aggregate_parameter` function I will transfer this line to the `aggregate_tremor` function

src/paradigma/tremor/tremor_analysis.py
nienketimmermans: I moved the function to `util.py`, but I did not add the mode yet, because this is still tremor-specific due to the bins.

Erikpostt: Is your method any different than simply extracting the most prevalent case?

nienketimmermans: Yes, it's a continuous variable, so I have to specifiy bins first

nienketimmermans:
@Erikpostt I added your suggestions and also changed the quantification step to aggregation step (and updated the test data accordingly). I also save the predictions for 'arm at rest' now, as I think this will improve understanding of why and how it is used (in both the prediction step and also later in the aggregation step). In a later stage when we are combining more function we could also think about how to add the calculation of the mode to the `aggregate_parameter` function.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,999 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 3: Detect tremor"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A pretrained logistic regression classifier is used to predict the tremor probability (`pred_tremor_proba`) for each window, based on the MFCCs. Using the prespecified threshold, a tremor label of 0 (no tremor) or 1 (tremor) is assigned (`pred_tremor_logreg`). Next, two extra criteria for rest tremor are applied. First, the frequency of the peak should be between 3-7 Hz. Second, the arm should be in rest or stable posture, defined as low frequency power below `config.movement_threshold` (established by splitting the low frequency power values measured during real-life into two clusters). The final tremor label is saved in `pred_tremor_checked`. A label for predicted arm at rest was also saved, to control for the amount of arm movement during the observed time period when aggregating the amount of tremor time in Step 4 (if a person is moving their arm, the cannot have rest tremor)."
```
Erikpostt: > the cannot have rest tremor

Small typo.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,999 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "A threshold of 50 deg²/s² is used to determine whether the arm is at rest or in stable posture.\n"
```
Erikpostt, line 761: Is deg^2/s^2 the unit for power you use?

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,999 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from paradigma.config import TremorFeatureExtractionConfig\n",
+    "from paradigma.tremor.tremor_analysis import extract_tremor_features\n",
+    "\n",
+    "config = TremorFeatureExtractionConfig()\n",
+    "print('The window length is', config.window_length_s, 'seconds.')\n",
```
Erikpostt: It is often cleaner to use fprinting:

`print(f'The window length is {config.window_length_s} seconds')`.

docs/tutorials/tremor_analysis.ipynb
Erikpostt: So `pred_arm_at_rest` is 1 if above threshold and 0 if below? I think this is the only column you don't specifically mention in the text. Also, perhaps you want to show `pred_tremor_checked` as the last one when showing the dataframe.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,999 @@
...
+   "source": [
+    "## Step 4: Compute aggregated tremor measures"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The final step is to compute the amount of tremor time and tremor power, aggregated over all windows in the input dataframe.\n",
+    "Tremor time is calculated as the number of the detected tremor windows, as percentage of the number of windows \n",
```
Erikpostt: I think this should be "the number of tremor windows"

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -0,0 +1,999 @@
...
+    "while the arm is at rest or in stable posture (when the low frequency power is below `config.movement_threshold`). This way the tremor time is controlled for the amount of time the arm is at rest or in stable posture, when rest tremor and re-emergent tremor could occur. For tremor power the following aggregates are derived: the median, mode and 90th percentile (specified in `config.percentile_tremor_power`) of tremor power. The median and modal tremor power reflect the typical tremor severity, whereas the 90th percentile reflects the maximal tremor severity within the observed timeframe. The aggregated tremor measures and metadata are stored in a json file."
```
Erikpostt: I don't see where the aggregates measures are stored. Did you miss this implementation?

Erikpostt: I think you removed `config.percentile_tremor_power` in this PR, and added a list `config.aggregates_tremor_power`.

Erikpostt (commented):
@nienketimmermans I've made some final suggestions.

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Yes, later it's converted to log scale

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: No `pred_arm_at_rest` is 1 if below the threshold and 0 if above. If the arm is moving, there is more power.

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Yes sharp

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Yes, forgot the adjust it here

nienketimmermans: Oh yes I don't actually store them anywhere. Do we want to add that in the tutorial?

nienketimmermans:
@Erikpostt I added the new suggestions. Do we want to save the aggregates file to some path in the tutorial? Then I will also implement that.

Erikpostt:
> @Erikpostt I added the new suggestions. Do we want to save the aggregates file to some path in the tutorial? Then I will also implement that.

I don't think we have to at this point in time, but we can add it later if desired. 

You can re-request my review and then I'll merge it.

Erikpostt (approved):
Ready to merge.

Issue #98, opened 2025-01-09
Classification parameters are currently stored inefficiently


Erikpostt:
As a user, I want to load the classifier object, classification threshold, and scaler parameters using a simple function instead of having the load _txt_ or _json_ files.

PR #99, opened 2025-01-09
Create classifier objects

Erikpostt:
The purpose of this PR is to solve #98. By creating a Class object we can include the classifier, classification threshold and scaling parameters in a single _pickle_ file. 

The following additional changes have been made:
- Change `os.path` to `Path` from `pathlib` to ensure robustness across platforms. Some operating systems use forward slash (/), others backward (\), and some double (//). This can create issues when working with `os`, while `Path` is considered more robust these days.
- Remove threshold and scaler files as these are now incorporated in the pickle file.
- Reduce memory used in classification by using arrays instead of column asignment.
- Bumping Python to `3.11.0`


Erikpostt:
@nienketimmermans @KarsVeldkamp I'm assigning both of you so you can inspect the suggestions for storing classifier objects. Let me know what you think about this method.

docs/notebooks/gait/gait_analysis.ipynb
```diff
@@ -35,27 +35,18 @@
     "branch = 'gait'\n",
     "sensor = 'imu'\n",
     "\n",
-    "gait_detection_classifier_filename = 'gait_detection_classifier.pkl'\n",
-    "gait_detection_scaler_filename = 'gait_detection_scaler_params.json'\n",
-    "gait_detection_threshold_filename = 'gait_detection_threshold.txt'\n",
+    "gait_detection_classifier_package_filename = 'gait_detection_clf_package.pkl'\n",
+    "gait_filtering_classifier_package_filename = 'gait_filtering_clf_package.pkl'\n",
```
KarsVeldkamp, line 39: nice one, clf_package instead of just clf (which I have currently in my pipeline)

nienketimmermans:
@Erikpostt looking good!

nienketimmermans (approved):

src/paradigma/gait/gait_analysis.py
```diff
@@ -172,40 +168,24 @@ def detect_gait(
...
         clf.n_jobs = 1
 
-    # Load and apply scaling parameters
-    with open(full_path_to_scaler, 'r') as f:
-        scaler_params = json.load(f)
-
-    scaler = StandardScaler()
-    scaler.mean_ = scaler_params['mean']
-    scaler.var_ = scaler_params['var']
-    scaler.scale_ = scaler_params['scale']
```
KarsVeldkamp, line 196: Maybe off-topic but: what is this scale actually? Since you also have the mu and sigma

KarsVeldkamp (approved):
Nice work!

src/paradigma/gait/gait_analysis.py
Erikpostt: Good question. You can check the documentation of `StandardScaler` [here](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html). When applying the `transform` function of the `StandardScaler`, it uses both the `'mean'` and `'scale'` parameters (see [here](https://github.com/scikit-learn/scikit-learn/blob/f159b78dc/sklearn/preprocessing/_data.py#L1044)). So, actually, the `'var'` parameter may be redundant.

Erikpostt:
@KarsVeldkamp @nienketimmermans thank you for reviewing, I'll merge this into main now. Be sure to check how this may affect your pipelines; you may have to create this classifier package by instantiating an object, loading the classifier and threshold (and optionally scaling parameters), adding these as parameters to the package, and storing the package object.

src/paradigma/gait/gait_analysis.py
KarsVeldkamp: Apologies, I think scale is the sigma (since it's basically np.sqrt(var_)).

PR #104, opened 2025-01-10
add tremor classifier package

nienketimmermans:
I created the classifier package for tremor detection and used Path instead of os


src/paradigma/tremor/tremor_analysis.py
```diff
@@ -287,5 +285,5 @@ def aggregate_tremor_io(path_to_feature_input: Union[str, Path], path_to_predict
     d_aggregates = aggregate_tremor(df, config)
 
     # Save output
-    with open(os.path.join(output_path,"tremor_aggregates.json"), 'w') as json_file:
+    with open(Path(output_path,"tremor_aggregates.json"), 'w') as json_file:
```
Erikpostt: When you use `Path`, you can simply define `output_path` as `Path('some_string')` and any child paths as `outputh_path / child / child`.

Erikpostt (approved):
Looking good, ready to merge. Feel free to change your usage of `Path` if you feel like it.

PR #105, opened 2025-01-10
Improve gait pipeline

Erikpostt:
Some minor changes to the gait pipeline. @KarsVeldkamp this is based on our discussion earlier today about saving a dataframe instead of a dict. This requires me to combine the forward and backward swings, otherwise the arrays are of unequal size, due to which I can't create a dataframe.


Erikpostt:
Ready for review.

@KarsVeldkamp do you understand why the order of the `time` and `values` in the meta files swap around? This keeps happening to me.

src/paradigma/gait/gait_analysis.py
```diff
@@ -260,6 +258,9 @@ def extract_arm_activity_features(
         A DataFrame containing the extracted arm activity features, including angle, velocity, 
         temporal, and spectral features.
     """
+    if sum(df_predictions[DataColumns.PRED_GAIT_PROBA] >= threshold) == 0:
```
KarsVeldkamp: Is this more efficient than if not any(df_predicitions...)

src/paradigma/gait/gait_analysis.py
```diff
@@ -526,12 +531,31 @@ def quantify_arm_swing(
...
     Returns
     -------
-    dict
-        A dictionary containing arm swing parameters for filtered and unfiltered gait, and per
-        segment length category.
+    Tuple[dict, dict]
+        A tuple containing a dictionary with quantified arm swing parameters for dfs_to_quantify, 
+        and a dictionary containing metadata for each segment.
     """
+    if sum(df_predictions[DataColumns.PRED_NO_OTHER_ARM_ACTIVITY_PROBA] >= classification_threshold) == 0:
```
KarsVeldkamp: See above

src/paradigma/gait/gait_analysis.py
```diff
@@ -526,12 +531,31 @@ def quantify_arm_swing(
...
+        raise ValueError("No gait without other arm activity detected in the input data.")
+    
+    if isinstance(dfs_to_quantify, str):
+        dfs_to_quantify = [dfs_to_quantify]
+    elif not isinstance(dfs_to_quantify, list):
+        raise ValueError("dfs_to_quantify must be either 'unfiltered', 'filtered', or a list containing both.")
```
KarsVeldkamp, line 550: Why can you end here with elif while in the generate_statistics this was an issue?

src/paradigma/gait/gait_analysis.py
```diff
@@ -617,116 +649,63 @@ def quantify_arm_swing(
...
+                    arm_swing_quantified[df_name].append(df_params_segment)
 
-            segment_results_aggregated[df_name][segment_cat] = cat_results
+        arm_swing_quantified[df_name] = pd.concat(arm_swing_quantified[df_name], ignore_index=True)
             
-    return segment_results_aggregated
+    return {df_name: arm_swing_quantified[df_name] for df_name in dfs_to_quantify}, segment_meta
 
 
-def quantify_arm_swing_io(
```
KarsVeldkamp, line 687: Did you delete the io wrapper entirely on purpose?

src/paradigma/util.py
```diff
@@ -5,7 +5,7 @@
 from pathlib import Path
 from datetime import timedelta
 from dateutil import parser
-from typing import List, Tuple
+from typing import List, Tuple, Callable, Any
```
KarsVeldkamp: Why do you insert these here?

KarsVeldkamp (approved):
Looking good! I posed a couple of questions and feel free to adjust and merge whenever you ready ;)

src/paradigma/gait/gait_analysis.py
Erikpostt: You're right, Ill adjust.

src/paradigma/gait/gait_analysis.py
Erikpostt: Because the function does not end here, so it still reaches the return statement. In the case of `generate_statistics` after the final `elif` there was nothing.

src/paradigma/gait/gait_analysis.py
Erikpostt: Yes, for this one I did. The main reason is that I removed _quantification_ and _aggregation_ from testing, and therefore no long require the `io` wrappers (since we use object assignment in the tutorials). The reason I removed these steps is because no arm swing was detected in the test data.

src/paradigma/util.py
Erikpostt: They were required for an intermediary solution, but very much redundant now. Oops.

PR #106, opened 2025-01-13
Tutorial ppg

KarsVeldkamp:
In this PR I created the tutorial for PPG but also did the following:
1. Changed the PPG preprocessing function (within preprocessing.py) to have a seperate io wrapper so I could load the tsdf data in memory before preprocessing. I did not include the loading function in the util.py while the other tsdf loading function is. Now I call it directly from tsdf --> what do you think of this?
2. Changed the ppg_analysis.ipynb accordingly
3. Changed from os.path.join to Path


docs/notebooks/ppg/ppg_analysis.ipynb
```diff
@@ -33,18 +35,42 @@
...
+   "source": [
+    "## Load the data"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 47,
+   "metadata": {},
+   "outputs": [],
+   "source": [
```
Erikpostt, line 71: If you don't use, e.g., `metadata_time`, you can call the function as `df_ppg, _, _ = ...` such that there are no objects assigned that hold memory.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   "metadata": {},
+   "source": [
+    "# Heart rate analysis"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This tutorial shows how to extract heart rate estimates using photoplethysmography (PPG) data and accelerometer data. The pipeline consists of a stepwise approach ot determine signal quality, assessing both PPG morphology and accounting for periodic artifacts using the accelerometer. Based on the signal quality, we extract high-quality segments and estimate the heart rate for every 2 s using the smoothed pseudo Wigner-Ville Distribution. "
```
Erikpostt: ot = to

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Load example data"
```
Erikpostt: Personally I would name this as 'Load data', and then in markdown we explain that we loaded our own data. That the user should load their data using their own methods.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Example PPG + accelerometer (from IMU files) data from a participant of the Personalized Parkinson Project is loaded. The data was prepared as explained in `data_preparation.ipynb`. The prepared IMU data contains both accelerometer and gyroscope data, but only accelerometer is used in this pipeline alongside the PPG data. We load the corresponding dataframes using the `load_tsdf_dataframe function`."
```
Erikpostt: Therefore, I would not explain what data is contained and what data is loaded. I'd simply specify that accelerometer and PPG data is required to run the PPG pipeline, and load that data here.

Erikpostt: I even think we should not mention `tsdf`, and perhaps even store a dataframe format (e.g., parquet, pickle or csv) of the prepared data used in this tutorial. I'll look into this.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+    "path_to_data = Path('../../tests/data')\n",
+    "path_to_prepared_data = path_to_data / '1.prepared_data'\n",
+    "\n",
+    "ppg_prefix = 'PPG'\n",
+    "imu_prefix = 'IMU'\n",
+    "\n",
+    "df_ppg, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data / ppg_prefix, prefix = ppg_prefix)\n",
+    "df_imu, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data / imu_prefix, prefix = imu_prefix)\n",
+    "\n",
+    "display(df_ppg, df_imu)"
```
Erikpostt, line 335: Do users know what "green" means?

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 1: Preprocess data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The first step after loading the data is to preprocess the data. Preprocessing starts by extracting the data where there is both PPG and IMU data. In this way, we discard the the first or last part of the segment when e.g. the PPG segment is recorded longer than the accelerometer. After this step, the preprocess_ppg_data function  resamples the values of both PPG and accelerometer data using uniformly distributed timestamps, since the sampling rate of both sensors is fixed but not uniform. After this, a bandpass filter (butterworth, 4th-order, cut-off frequencies: [0.4, 3.5]) is applied to the PPG signal and a high-pass (butterworth, 4th-order, cut-off: 0.2 Hz) filter is applied to the accelerometer. "
```
Erikpostt: Here's how I would write it, feel free to merge freely:

> The first step after loading the data is preprocessing. This begins by isolating segments containing both PPG and IMU data, discarding portions where one modality (e.g., PPG) extends beyond the other, such as when the PPG recording is longer than the accelerometer data. After this step, the `preprocess_ppg_data` function resamples the PPG and accelerometer data to uniformly distributed timestamps, addressing the fixed but non-uniform sampling rates of the sensors. After this, a bandpass Butterworth filter (4th-order, bandpass frequencies: 0.4--3.5 Hz) is applied to the PPG signal, while a high-pass Butterworth filter (4th-order, cut-off frequency: 0.2 Hz) is applied to the accelerometer data.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Shape of the original data: (64775, 2) (72947, 4)\n",
```
Erikpostt: I think this is difficult to interpret for some users, so perhaps you could clarify rows x columns, and refer to the corresponding dataframes.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+      "Shape of the overlapping segments: (64775, 2) (64361, 4)\n"
```
Erikpostt: Why are the shapes different? I guess this is because of the sampling frequency, but might be useful to clarify.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 2: Extract signal quality features"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The preprocessed data (PPG & accelerometer) is windowed into overlapping windows of length `ppg_config.window_length_s` with a window step of `ppg_config.window_step_length_s`. From the PPG windows 10 time- and frequency domain features are extracted to assess PPG morphology and from the accelerometer windows one relative power feature is calculated to assess periodic motion artifacts. "
```
Erikpostt, line 680: It's useful to specify the value here by printing `ppg_config.window_length_s` for example.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 3: Signal quality classification"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "A trained logistic classifier is used to classify PPG signal quality and returns the `pred_sqa_proba`, which is the posterior probability of a PPG window to look like the typical PPG morphology (higher probability indicates toward the typical PPG morphology). The relative power feature from the accelerometer is compared to a threshold for periodic artifacts and therefore `pred_sqa_acc_label` returns a label indicating probably periodic motion artifacts (label 0) or no periodic motion artifacts (label 1). "
```
Erikpostt: "A trained logistic classifier is used to predict the PPG signal quality `pred_sqa_proba`: the posterior probability of a PPG window ..."

I'm having slight difficulties interpreting this term: does it return a probability of a signal quality (my interpretation of this text), or does it return a signal quality float (my interpretation of the code)?

Erikpostt: & I would remove "probably", perhaps make it "predicted".

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+     "execution_count": 28,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "from paradigma.config import SignalQualityClassificationConfig\n",
+    "from paradigma.heart_rate.heart_rate_analysis import signal_quality_classification\n",
+    "\n",
+    "path_to_classifier = Path('../../tests/data/0.Classification/ppg')\n",
```
Erikpostt, line 1096: The classifier path should be in `assets`, but I'm sure you'll do this with your next PR on classification.

Erikpostt: It might be worthwhile to print the classification threshold here, such that users can also modify it to their liking.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 4: Heart rate estimation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For heart rate estimation, we extract segments of `config.tfd_length`. Using a 2-step approach, we calculate the smoothed-pseudo Wigner-Ville Distribution to obtain the frequency content of the PPG signal over time. For every 2-second window, we identified the frequency with the highest power for each data point and assigned the average of these frequency as the heart rate in that 2-second window. "
```
Erikpostt: Also print the tsdf_length here.

Erikpostt: What's the 2-step approach? Does it add value to mention it here?

Erikpostt: Check the tense of your language: present vs past.

Erikpostt: With data point you mean each timestamp constituting the 2-second window, right? How does each timestamp have a frequency spectrum? Does that have to do with the WVD?

And perhaps, could you explain what the relation is between "the average of these frequencies" and "the heart rate in that 2-second window"? I can't fully make that equation, might be useful to clarify how A represents B.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+       "\n",
+       "    .dataframe thead th {\n",
+       "        text-align: right;\n",
+       "    }\n",
+       "</style>\n",
+       "<table border=\"1\" class=\"dataframe\">\n",
+       "  <thead>\n",
+       "    <tr style=\"text-align: right;\">\n",
+       "      <th></th>\n",
+       "      <th>rel_time</th>\n",
```
Erikpostt: What's relative time? Could you explain why this changed from `time` to `rel_time`?

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   "metadata": {},
+   "source": [
+    "## Step 5: Heart rate aggregation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The final step is to aggregate all 2-second heart rate estimates. In the current example, the mode and 99th percentile are calculated. The mode represent the resting heart rate while the 99th percentile indicates the maximum heart rate. In Parkinson's disease, we expect that these two measures could reflect autonomic (dys)functioning."
```
Erikpostt: How is the mode representing the resting heart rate? Is this because the most prevalent heart rate is when we are resting?

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -0,0 +1,1245 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "{'hr_aggregates': {'99p_heart_rate': 86.62868369351669,\n",
+      "                   'mode_heart_rate': 84.8722986247544},\n",
+      " 'metadata': {'nr_hr_est': 6}}\n"
```
Erikpostt, line 1243: `nr_hr_est` is difficult for me to interpret, could you explain?

src/paradigma/preprocessing.py
Erikpostt, line 1: If people collect both accelerometer and PPG data, longitudinally, this function actually might make sense to have as a utility in the toolbox. I would move it to `util.py` therefore. I would leave all "necessary" functionalities in `preprocessing.py`.

src/paradigma/preprocessing.py
```diff
@@ -245,6 +245,21 @@ def preprocess_imu_data_io(path_to_input: str | Path, path_to_output: str | Path
 
 
 def scan_and_sync_segments(input_path_ppg, input_path_imu):
```
Erikpostt: Please set strict data types for the input parameters.

src/paradigma/preprocessing.py
Erikpostt: I do think this functionality is `tsdf`-specific, so we should think about how a user using `csv` or other formats could do this.

src/paradigma/preprocessing.py
```diff
@@ -350,25 +355,62 @@ def preprocess_ppg_data(tsdf_meta_ppg: tsdf.TSDFMetadata, tsdf_meta_imu: tsdf.TS
...
+
+    Parameters
+    ----------
+    tsdf_meta_ppg : tsdf.TSDFMetadata
+        Metadata for the PPG data.
+    tsdf_meta_imu : tsdf.TSDFMetadata
+        Metadata for the IMU data.
+    output_path : Union[str, Path]
+        Path to store the preprocessed data.
+    ppg_config : PPGPreprocessingConfig
```
Erikpostt: PPGConfig

src/paradigma/preprocessing.py
```diff
@@ -350,25 +355,62 @@ def preprocess_ppg_data(tsdf_meta_ppg: tsdf.TSDFMetadata, tsdf_meta_imu: tsdf.TS
...
+        Configuration object for PPG preprocessing.
```
Erikpostt, line 376: IMUConfig

src/paradigma/preprocessing.py
```diff
@@ -350,25 +355,62 @@ def preprocess_ppg_data(tsdf_meta_ppg: tsdf.TSDFMetadata, tsdf_meta_imu: tsdf.TS
...
+    imu_config : IMUPreprocessingConfig
+        Configuration object for IMU preprocessing.
+    sensor: str
```
Erikpostt: This parameter is removed.

Erikpostt (changes requested):
I have raised some questions and made some suggestions. Let me know what you think.

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: Agree, I have adjusted it now and made it more concise.

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: Good point, I have added a brief sentence to the documentation

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: Thanks, I like the suggestion!

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: Yes good point, which I have added now

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: good point

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: agree

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: @Erikpostt I am not sure if I understand you correctly but both are returning a probability of a signal quality right? And not just a float?

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: Yes I will modify this in the next PR and this will be for the prerelease hopefully as well

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: To comment on your last questions: 
1. Yes WVD returns a frequency spectrum for every data point (because it uses every lag in the autocorrelation function)
2. We extract for every timestamp in that 2s window the frequency with the most power. This results in an array of "dominant" frequencies which I average.

KarsVeldkamp: I have tried to make this more self-explaining

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: Nothing, I'll adjust it to time ;)

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: yes

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: yes

src/paradigma/preprocessing.py
KarsVeldkamp: I am not sure where you exactly referring to but I assume this is to scan_and_sync_segments? This is on my list to remove it from preprocessing and set it to util or another place and it is tsdf-specific indeed

src/paradigma/preprocessing.py
KarsVeldkamp: THis is very old documentation ;)

src/paradigma/preprocessing.py
KarsVeldkamp: Then also the extract_meta_from_tsdf_files will be removed in the preprocessing.py

KarsVeldkamp:
@Erikpostt I have made changes based on your feedback (visible in commit [40e768a](https://github.com/biomarkersParkinson/paradigma/pull/106/commits/40e768af04617e3e89f7b6300bb7460b3e65c263). Let me know what you think of this

docs/tutorials/heart_rate_analysis.ipynb
Erikpostt: Is the signal quality binary, or not? The first (probability of a signal quality) can be seen as similar to gait: it's either gait or non-gait, and there's a probability attached to it (in this case, probability that it is of high quality). The second, signal quality is always a float, and signal quality probability is basically just signal quality (as it is a float). Does this make more sense?

src/paradigma/preprocessing.py
Erikpostt: Yes apologies, it's about `scan_and_sync_segments`. Check.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -346,7 +346,9 @@
...
+    "The first step after loading the data is preprocessing. This begins by isolating segments containing both PPG and IMU data, discarding portions where one modality (e.g., PPG) extends beyond the other, such as when the PPG recording is longer than the accelerometer data. After this step, the preprocess_ppg_data function resamples the PPG and accelerometer data to uniformly distributed timestamps, addressing the fixed but non-uniform sampling rates of the sensors. After this, a bandpass Butterworth filter (4th-order, bandpass frequencies: 0.4--3.5 Hz) is applied to the PPG signal, while a high-pass Butterworth filter (4th-order, cut-off frequency: 0.2 Hz) is applied to the accelerometer data.\n",
+    "\n",
+    "Note: the printed shapes are the data points (rows) and the data columns (rows x columns). The number of rows of the overlapping segments of PPG and accelerometer are not exactly the same due to sampling differences (other sensors and possibly other sampling frequencies). "
```
Erikpostt: I think you meant to say (rows, columns), with each row corresponding to a single data points? Now it looks like rows are data points, and rows x columns are data columns.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -953,7 +968,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "A trained logistic classifier is used to classify PPG signal quality and returns the `pred_sqa_proba`, which is the posterior probability of a PPG window to look like the typical PPG morphology (higher probability indicates toward the typical PPG morphology). The relative power feature from the accelerometer is compared to a threshold for periodic artifacts and therefore `pred_sqa_acc_label` returns a label indicating probably periodic motion artifacts (label 0) or no periodic motion artifacts (label 1). "
+    "A trained logistic classifier is used to predicts PPG signal quality and returns the `pred_sqa_proba`, which is the posterior probability of a PPG window to look like the typical PPG morphology (higher probability indicates toward the typical PPG morphology). The relative power feature from the accelerometer is compared to a threshold for periodic artifacts and therefore `pred_sqa_acc_label` returns a label indicating predicted periodic motion artifacts (label 0) or no periodic motion artifacts (label 1). "
```
Erikpostt: predicts = predict
returns = return (as in, "is used to return", but could also be "returns" as in "and it returns")

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -1090,14 +1105,21 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "For heart rate estimation, we extract segments of `config.tfd_length`. Using a 2-step approach, we calculate the smoothed-pseudo Wigner-Ville Distribution to obtain the frequency content of the PPG signal over time. For every 2-second window, we identified the frequency with the highest power for each data point and assigned the average of these frequency as the heart rate in that 2-second window. "
+    "For heart rate estimation, we extract segments of `config.tfd_length`. We calculate the smoothed-pseudo Wigner-Ville Distribution (SPWVD) to obtain the frequency content of the PPG signal over time. We extract for every timestamp in the SPWVD for the frequency with the highest power. For every non-overlapping 2-second window we average the corresponding frequencies to obtain a heart rate per window. "
```
Erikpostt: "... SPWVD for the ..." should be "... SPWVD the ..."

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -1193,7 +1216,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The final step is to aggregate all 2-second heart rate estimates. In the current example, the mode and 99th percentile are calculated. The mode represent the resting heart rate while the 99th percentile indicates the maximum heart rate. In Parkinson's disease, we expect that these two measures could reflect autonomic (dys)functioning."
+    "The final step is to aggregate all 2-second heart rate estimates. In the current example, the mode and 99th percentile are calculated. We hypothesize that the mode gives representation of the resting heart rate while the 99th percentile indicates the maximum heart rate. In Parkinson's disease, we expect that these two measures could reflect autonomic (dys)functioning. The `nr_hr_est` in the metadata indicates on how many 2-second windows these aggregates are determined."
```
Erikpostt: Not sure, but should "... indicates on how many ..." be "... indicates based on how many ..."?

Also try to be consistent in using "2-second" or "2 s". I think I saw "2 s" somewhere.

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -927,17 +941,18 @@
...
      "output_type": "execute_result"
     }
    ],
    "source": [
     "from paradigma.config import SignalQualityFeatureExtractionConfig, SignalQualityFeatureExtractionAccConfig\n",
     "from paradigma.heart_rate.heart_rate_analysis import extract_signal_quality_features\n",
-    "\n",
     "ppg_config = SignalQualityFeatureExtractionConfig()\n",
     "acc_config = SignalQualityFeatureExtractionAccConfig()\n",
+    "print(\"The default window length for the signal quality feature extraction is set to\", ppg_config.window_length_s, \"seconds.\")\n",
```
Erikpostt, line 957: This cell is a bit cluttered. Try to add some whitespaces 

- imports
- configs
- prints
- function
- print

Erikpostt: This also applies to some other cells.

Erikpostt (approved):
I've made some small suggestions, otherwise ready to merge.

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp: Yes I made slight adjustments in this sentence

KarsVeldkamp:
@Erikpostt Thanks for your feedback, I have made the last adjustments and will merge it now!

PR #108, opened 2025-01-13
Total psd correction

nienketimmermans:
@Erikpostt I fixed the total psd calculation and added function links in the tutorial.


docs/tutorials/tremor_analysis.ipynb
```diff
@@ -219,11 +219,10 @@
     "from paradigma.util import load_tsdf_dataframe\n",
     "\n",
     "path_to_data =  Path('../../tests/data')\n",
-    "path_to_assets = Path('../../src/paradigma/assets')\n",
     "path_to_prepared_data = path_to_data / '1.prepared_data' / 'imu'\n",
-    "df_data, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data, prefix='IMU')\n",
+    "df_data, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data, prefix='IMU') # load the data (saved in TSDF format)\n",
```
Erikpostt: I think your line is too long. In this case, it is best practice to have the comment *above*  the code.

In general, comments above the code is suggested (partly because of this reason), although in-line comments can be useful in some cases (see [here](https://peps.python.org/pep-0008/#comments)).

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -219,11 +219,10 @@
...
     "\n",
-    "df_data"
+    "df_data # print the dataframe"
```
Erikpostt: I would not comment this, as this is quite obvious (again, see [here](https://peps.python.org/pep-0008/#comments) for best practices).

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -219,11 +219,10 @@
     "from paradigma.util import load_tsdf_dataframe\n",
     "\n",
     "path_to_data =  Path('../../tests/data')\n",
```
Erikpostt, line 222: I would clarify here that the user should set the path here if desired.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -944,7 +947,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The final step is to compute the amount of tremor time and tremor power, aggregated over all windows in the input dataframe.\n",
+    "The final step is to compute the amount of tremor time and tremor power with the function [aggregate_tremor](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/tremor/tremor_analysis.py#:~:text=aggregate_tremor), which aggregates over all windows in the input dataframe.\n",
```
Erikpostt, line 951: You might want to print the parameters of config here, such that the user knows how to select them.

src/paradigma/tremor/feature_extraction.py
```diff
@@ -286,8 +286,8 @@ def extract_spectral_domain_features(config, data) -> pd.DataFrame:
         axis=1
     )
 
-    # Compute total power in the PSD
-    total_psd = compute_total_power(psd)
+    # Compute total power in the PSD (sum over the three axes)
+    total_psd = np.sum(psd, axis=2)
```
Erikpostt: You can use `compute_total_power` in _gait.feature_extraction_ which should do the trick too. If you change this, I'll make sure it functions when I move `compute_total_power` to the general _feature_extraction_ file.

Erikpostt (changes requested):
I have made several suggestions for changes, please take a look and let me know what you think.

src/paradigma/tremor/feature_extraction.py
nienketimmermans: @Erikpostt that function sums over the frequency bins right? Not over the axes?

Erikpostt: The `compute_total_power` in [gait](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/gait/feature_extraction.py#L126) has `axis=-1`, meaning the first from the right, which is equal to `axis=2` for a 3D array. So, it should be similar to your code. The difference is that `axis=2` does not work for 2D arrays, while `axis=-1` does.

nienketimmermans:
Thanks for the comments about the docstrings, I've adjusted them. 
I didn't adjust the total psd calculation, as this function is summing over frequency bins instead of the axes. Should I name it summed_psd instead of total_psd to avoid confusion?

src/paradigma/tremor/feature_extraction.py
nienketimmermans: Ah that explains why I got the same results. But we might want to adjust the documentation of the function then to avoid confusion.

nienketimmermans:
Thanks for the comments about the docstrings, I've adjusted them. I also changed the total_psd calculation again to how it was. I'm not sure why you got a 1D output there.

src/paradigma/tremor/feature_extraction.py
Erikpostt: You're right, I'll include this is an upcoming PR.

docs/tutorials/tremor_analysis.ipynb
Erikpostt: It's a step forward, though as a user I might be confused about my data not being in TSDF.  You might want to clarify that the test data is stored in TSDF (note that the "F" means "format"), and that the user can load their own data in their own way.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -961,6 +962,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
+      "Besides the modal tremor power the following tremor power aggregates are derived: ['median', '90p'].\n",
```
Erikpostt: So you always compute the modal, and other parameters are optional? Perhaps you could also make the modal optional, as a user might not necessary request this specific aggregate.

Erikpostt (approved):
Looking good. I've made several more suggestions, but otherwise ready to merge.

PR #110, opened 2025-01-14
Merge feature files

Erikpostt:
Requesting review of both @KarsVeldkamp and @nienketimmermans as this has implications for the entire module.

This is an extensive PR intended to reduce the complexity of the directory structure. I suggest the following changes:
- Change __analysis.py_ files to __pipeline.py_ for more accurate naming convention;
- Move _pipeline_ files to a separate subdirectory;
- Combine _feature_extraction_ files into a single file in the _src_ directory (a file _heart_rate_utils.py_ is added to the _pipelines_ directory to include TFD-related functionalities, which I converted to a Class);
- Remove subdirectories in _notebooks_ directory;

Also added to this PR are:
- Re-instantiating the `io` wrappers for testing notebooks (we have to think about how to create new test data when the pipelines are modified in a more sustainable manner, and in a way hidden from the user);
- Swap the order of parameters of functions for consistency;
- Move `merge_predictions_with_timestamps` to _util.py_ -> let's take a look later whether this is useful for other domains too

@nienketimmermans note that one of your _bin_ values changed, but your aggregates remained the same.


Erikpostt:
@KarsVeldkamp @nienketimmermans ready for review.

docs/tutorials/heart_rate_analysis.ipynb
KarsVeldkamp, line 1: Also here, the outputs are removed but maybe not that relevant since we have to update the notebooks with the new config and the new structure overall

src/paradigma/feature_extraction.py
```diff
@@ -227,25 +431,28 @@ def compute_mfccs(
...
+        of the signal for each window.
     config : object
         Configuration object containing the following attributes:
         - window_length_s : int
             Duration of each analysis window in seconds.
         - sampling_frequency : int
             Sampling frequency of the data (default: 100 Hz).
         - mfcc_low_frequency : float
-            Lower bound of the frequency band (default: 0 Hz).
+            Lower bound of the frequency band (default: 0).
```
KarsVeldkamp: Why is here the Hz removed while for the high_frequency it is there?

src/paradigma/util.py
```diff
@@ -378,3 +378,83 @@ def aggregate_parameter(parameter: np.ndarray, aggregate: str) -> np.ndarray:
         return np.std(parameter)
     else:
         raise ValueError(f"Invalid aggregation method: {aggregate}")
+
+def merge_predictions_with_timestamps(
```
KarsVeldkamp, line 382: This function is very similar to the assign_sqa_label in my heart_rate_utils.py --> maybe at a later stage we can think of how to combine these into one utility function

KarsVeldkamp (approved):
Great work @Erikpostt! Nice to see how this effort helps to clean up the toolbox. The only thing we have to do is to change the tutorials according to this PR

src/paradigma/util.py
Erikpostt: Yes I agree, I was already quite curious whether you had similar functionalities. Let's put it on the list.

src/paradigma/feature_extraction.py
```diff
@@ -80,47 +118,51 @@ def compute_power_in_bandwidth(config, psd: np.ndarray, freqs: np.ndarray) -> di
...
-          tuples of frequency ranges (low, high) in Hz.
-        - `sensor` (str): The name of the sensor used for prefixing column names.
-    psd : np.ndarray
-        A 3D array of shape (n_windows, n_frequencies, n_axes) representing the 
-        power spectral density (PSD) of the sensor data.
     freqs : np.ndarray
         A 1D array of shape (n_frequencies,) containing the frequencies corresponding 
         to the PSD values.
+    psd : np.ndarray
+        A 2D array of shape (n_windows, n_frequencieS) or 3D array of shape (n_windows, n_frequencies, n_axes)
```
nienketimmermans: There is a small typo here (n_frequencie**S**)

src/paradigma/feature_extraction.py
```diff
@@ -208,15 +302,125 @@ def compute_dominant_frequency(
...
+    ----------
+    freqs: np.ndarray
+        The frequency bins of the power spectral density.
+    psd: np.ndarray
+        The power spectral density of the signal.
+    config: SignalQualityFeatureExtractionConfig
+        The configuration object containing the parameters for the feature extraction. The following
+        attributes are used:
+        - freq_band_physio: tuple
+            The frequency band for physiological tremor (default: (0.75, 3)).
```
nienketimmermans: I think you mean HR instead of tremor?

src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -286,3 +289,83 @@ def aggregate_tremor_io(path_to_feature_input: Union[str, Path], path_to_predict
...
     with open(output_path / "tremor_aggregates.json", 'w') as json_file:
         json.dump(d_aggregates, json_file, indent=4)
+
+
+def extract_spectral_domain_features(data: np.ndarray, config) -> pd.DataFrame:
+    """
+    Compute spectral domain features from the gyroscope data.
+
+    This function computes Mel-frequency cepstral coefficients (MFCCs), the frequency of the peak, 
+    the tremor power, and the non-tremor power based on the total power spectral density of the windowed gyroscope data.
```
nienketimmermans: Could you change non-tremor power to low frequency power? I see that I also save the variable in the dictionary as that

nienketimmermans (approved):
@Erikpostt thank you for organizing all the files! I have posted a few comments, but it's ready to be merged after you've looked at them.

PR #112, opened 2025-01-16
Iterate on tutorial gait

Erikpostt:
A next iteration of the gait and data preparation tutorials.


docs/tutorials/data_preparation.ipynb
KarsVeldkamp, line 1: In general I have two remarks: 1) we discussed that the scale factors are TSDF specific and are maybe not needed to describe here because it can be confusing, 2) also we discussed that the watch side could be implemented in the pipeline itself and that you insert the watch side as parameter in your pipeline. 
I'll leave it up to you what you want to do with this now.

docs/tutorials/gait_analysis.ipynb
```diff
@@ -717,323 +619,859 @@
...
+    "\n",
+    "print(f\"A total of {df_gait.shape[1]-1} features have been extracted from {df_gait.shape[0]} {config.window_length_s}-second windows with {config.window_length_s-config.window_step_length_s} seconds overlap.\")\n",
+    "df_gait.head()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Each row in this dataframe corresponds to a single window, with the window length and overlap set in the `config` object. Note that the `time` column has a 1-second interval instead of the 10-millisecond interval before, as it now represents the starting time of the window relative to the first data point of the dataframe."
```
KarsVeldkamp: This is only in the data_preparation tutorial right that it was a delta instead of relative time. I get the point that you want to emphasize that you have the window steps of 1s isntead of data point times but they are both relative tot the first data point of the dataframe so I think you can leave that part out

docs/tutorials/gait_analysis.ipynb
```diff
@@ -717,323 +619,859 @@
...
+    "print(f\"A total of {df_arm_activity.shape[1]-1} features have been extracted from {df_arm_activity.shape[0]} {config.window_length_s}-second windows with {config.window_length_s-config.window_step_length_s} seconds overlap.\")\n",
+    "df_arm_activity.head()"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Each row in this dataframe corresponds to a single window, with the window length and overlap set in the `config` object. Note that the `time` column has a 1-second interval instead of the 10-millisecond interval before, as it now represents the starting time of the window relative to the first data point of the dataframe."
+    "The features extracted are similar to the features extract for gait detection, but the gyroscope has been added to extract additional MFCCs of this sensor. The gyroscope (measuring rotation) is relevant to distinguish between arm activities. Also note that the `time` column no longer starts at 0, since the first timestamps were predicted as non-gait and therefore discarded."
```
KarsVeldkamp: Technically gyroscope measures angular velocity right?

KarsVeldkamp (approved):
I have posed some questions and wrote down some minor remarks but leave it up to you what you want to do with them ;)

docs/tutorials/gait_analysis.ipynb
Erikpostt: Correct ;)

Erikpostt:
@KarsVeldkamp could you please check the new changes? I have done the following:

- Created a separate directory for tutorial data, which has the scaled data
- Inverting watch side added to preprocessing
- Tutorials modified according to your suggestions

src/paradigma/preprocessing.py
KarsVeldkamp, line 1: Yes! It was already a utility function btw?

KarsVeldkamp (approved):
Keep up the great work you're doing @Erikpostt!

src/paradigma/preprocessing.py
Erikpostt: Yup :)

PR #117, opened 2025-01-17
Pre release preparations

Erikpostt:
Making some final changes for the pre-release.
- Update docs
- Move notebooks to _test_ directory
- Change notebooks to `io` format: I assume here we want to use the notebooks solely for testing purposes, and the tutorial notebooks specifically for users to work with
- Move `io` functionalities to _testing.py_ files such that it does not clutter other files

P.S. I can't seem to restore the execution counts in the notebooks any longer. If we think this is too much of an issue I can look into it later.


Erikpostt:
Hey @KarsVeldkamp, one final PR before the pre-release I suspect.

docs/conf.py
```diff
@@ -7,9 +7,9 @@
 # -- Project information -----------------------------------------------------
 
 project = "paradigma"
-copyright = "2023, Peter Kok, Vedran Kasalica, Erik Post, Kars Veldkamp, Nienke Timmermans, Diogo Coutinho Soriano, Luc Evers"
+copyright = "2023, Erik Post, Kars Veldkamp, Nienke Timmermans, Peter Kok, Vedran Kasalica, Diogo Coutinho Soriano, Luc Evers"
 author = (
-    "Peter Kok, Vedran Kasalica, Erik Post, Kars Veldkamp, Nienke Timmermans, Diogo Coutinho Soriano, Luc Evers"
+    "Erik Post, Kars Veldkamp, Nienke Timmermans, Peter Kok, Vedran Kasalica, Diogo Coutinho Soriano, Luc Evers"
```
KarsVeldkamp, line 12: ;)

src/paradigma/preprocessing.py
```diff
@@ -282,9 +256,7 @@ def preprocess_ppg_data(df_ppg: pd.DataFrame, df_imu: pd.DataFrame, ppg_config:
     df_acc = df_imu.drop(cols_to_drop, axis=1)
 
     # Extract overlapping segments
-    print(f"Original data shapes:\n- PPG data: {df_ppg.shape}\n- Accelerometer data: {df_acc.shape}")
     df_ppg_overlapping, df_acc_overlapping = extract_overlapping_segments(df_ppg, df_acc, start_time_ppg, start_time_imu)
-    print(f"Overlapping data shapes:\n- PPG data: {df_ppg_overlapping.shape}\n- Accelerometer data: {df_acc_overlapping.shape}")
```
KarsVeldkamp, line 287: I printed this on purpose so a user knows what's happening to the dataframe. But we can argue whether that's useful

src/paradigma/testing.py
KarsVeldkamp, line 1: not all io's are fully implemented in my pipeline but that's not an issue for now I think

KarsVeldkamp (approved):
This is looking good, made a minor remark about the printing (because that also influences my tutorial but can adjust it easily). 
Otherwise, ready to merge!

src/paradigma/preprocessing.py
Erikpostt: The reason why I removed it here is because in general you don't want functions to have `print` statements. You could also print the shape of the dataframe before calling this function, and afterwards, to get an idea, right? So perhaps that could be added to the tutorial?

src/paradigma/testing.py
Erikpostt: That's okay, it's just for testing purposing.

PR #138, opened 2025-01-28
Adjust mfcc computation for tremor pipeline

nienketimmermans:
@Erikpostt I modified the mfcc computation and tremor classifier so that it uses the original classifier again. Let me know if you have other ideas for combining the compute_mfcc function for the tremor and gait pipelines


docs/tutorials/tremor_analysis.ipynb
```diff
@@ -961,10 +961,10 @@
      "output_type": "stream",
      "text": [
       "Besides the modal tremor power the following tremor power aggregates are derived: ['mode', 'median', '90p'].\n",
-      "{'aggregated_tremor_measures': {'90p_tremor_power': 1.796145868028196,\n",
-      "                                'median_tremor_power': 0.545156024401169,\n",
-      "                                'modal_tremor_power': 0.52,\n",
-      "                                'perc_windows_tremor': 11.235955056179774},\n",
+      "{'aggregated_tremor_measures': {'90p_tremor_power': 0.8001855655603451,\n",
```
Erikpostt, line 964: Are these aggregates correct? Is it such a large difference with the change in MFCCs?

src/paradigma/feature_extraction.py
```diff
@@ -469,6 +475,10 @@ def compute_mfccs(
     # Compute window length in samples
     window_length = config.window_length_s * config.sampling_frequency
     
+    # Determine the lenght of subwindows used in the spectrogram computation
```
Erikpostt: - Typo: _lenght_ => _length_
- Perhaps you could set `total_power_array.shape[2]` to a variable, such that users know what this variable is and how the window length is defined.

src/paradigma/feature_extraction.py
```diff
@@ -483,10 +493,16 @@ def compute_mfccs(
             config.mfcc_high_frequency, 
             num=config.mfcc_n_dct_filters + 2
         )
-
-    filter_points = np.floor(
-        window_length / config.sampling_frequency * freqs
-    ).astype(int)  + 1
+    
+    if pipeline == 'tremor':
+        filter_points = np.round(
```
Erikpostt, line 507: Why do you change it to `np.round`? Wasn't it `np.floor` before when we discussed how to align the tremor and gait pipelines?

src/paradigma/feature_extraction.py
Erikpostt, line 1: Here's my general opinion on your suggested changes. I think you found a good way to solve the problem, but I can imagine this is not always self-explanatory to the user. 

- Why did you opt for letting the user select a pipeline (_gait_ or _tremor_), instead of letting the user select a method (_spectrogram_ or _psd_)? 
- Why do you check the shape/dimensions as if statements instead of referring back to the parameter (in this case pipeline)?

Erikpostt (commented):
Good work, I've left some questions that would be great to discuss before merging the changes.

src/paradigma/assets/tremor_detection_clf_package.pkl
Erikpostt, line 1: Where did the classifier go? Or is it just reduced significantly because of the feature changes?

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Yes, the differences were smaller when aggregating over a whole week. Detecting only a few different tremor windows in 8 minutes of data led to larger differences.

src/paradigma/feature_extraction.py
nienketimmermans: No I used `np.round` before, and it leads to small differences

src/paradigma/feature_extraction.py
nienketimmermans: At first I did not include the option to select the pipeline to make it easier, so I used the shape of the input to determine the necessary steps. But then I found the difference between np.round and np.floor, which are not related to the choice of spectrogram or psd, but to choices that we made for the individual pipelines. What do you think is the best to do? Change all if statements where the shape is checked to check the pipeline parameter? Or change this to a method parameter (spectrogram or psd)?

src/paradigma/assets/tremor_detection_clf_package.pkl
nienketimmermans: Apparently it is reduced significantly, maybe because I specified the coefficients in a slightly different way now.

docs/tutorials/tremor_analysis.ipynb
Erikpostt: Ah that makes sense.

src/paradigma/feature_extraction.py
Erikpostt: Difficult one ... I think the statements should be both independent of a selected pipeline and of the shape. For `pipeline`, perhaps a user would like to use the spectrogram instead of the PSD, but then has to set `pipeline` to `gait`. So I believe a user should select either `'psd'` or `'spectrogram'` as parameter value, with perhaps a default value set such that it is not expected of a user per se. Similarly, can't the parameters `psd` and `spectrogram` both be 2D and 3D? We could also opt for a "rounding" parameter of sorts, which can be set to `'floor'` or `'round'` with one value set as default.

src/paradigma/assets/tremor_detection_clf_package.pkl
Erikpostt: You might want to check it briefly, but otherwise no issues from my side.

src/paradigma/feature_extraction.py
nienketimmermans: Yes, good idea. I will remove the pipeline specification. However, I don't think spectrogram can be 2D, as it is by definition the spectrum over time (so you have the frequency axis, the amplitude/power axis and the time axis).

nienketimmermans:
@Erikpostt I've now included the parameters `total_power_type` and `rounding_method` instead of `pipeline` in the `compute_mfcc` function, with the default set to `psd` and `floor`.

src/paradigma/feature_extraction.py
Erikpostt: Ah you're right

Erikpostt (approved):
Looking good, ready to merge. A brief suggestion (for now or for later): when you request strings to be selected among a list of options (e.g., for `total_power_type` it should be either `'psd'` or `'spectrogram'` , you could force the user to select either of the two by raising a `ValueError` if none of the two are selected as follows:

``` 
if total_power_type not in ['psd', 'spectrogram']:
   raise ValueError("total_power_type should be set to either 'psd' or 'spectrogram'")
```

PR #139, opened 2025-02-03
fix issue with no tremor

nienketimmermans:
If no tremor is detected, the aggregated tremor power measures are not derived and NaN is saved as output for these measures.

### DIFF
--- a/src/paradigma/pipelines/tremor_pipeline.py
+++ b/src/paradigma/pipelines/tremor_pipeline.py
@@ -183,36 +183,53 @@ def aggregate_tremor(df: pd.DataFrame, config: TremorConfig):
     # calculate tremor time
     perc_windows_tremor= np.sum(df_filtered['pred_tremor_checked']) / nr_windows_rest * 100 # as percentage of total measured time without non-tremor arm movement
 
-    # calculate aggregated tremor power measures
-    tremor_power = df_filtered.loc[df_filtered['pred_tremor_checked'] == 1, 'tremor_power']
-    tremor_power = np.log10(tremor_power+1) # convert to log scale
-    aggregated_tremor_power = {}
-    
-    for aggregate in config.aggregates_tremor_power:
-        aggregate_name = f"{aggregate}_tremor_power"
-        if aggregate == 'mode':
-            # calculate modal tremor power
-            bin_edges = np.linspace(0, 6, 301)
-            kde = gaussian_kde(tremor_power)
-            kde_values = kde(bin_edges)
-            max_index = np.argmax(kde_values)
-            aggregated_tremor_power['modal_tremor_power'] = bin_edges[max_index]
-        else: # calculate te other aggregates (e.g. median and 90th percentile) of tremor power
-            aggregated_tremor_power[aggregate_name] = aggregate_parameter(tremor_power, aggregate)
-    
-    # store aggregates in json format
-    d_aggregates = {
-        'metadata': {
-            'nr_windows_total': nr_windows_total,
-            'nr_windows_rest': nr_windows_rest
-        },
-        'aggregated_tremor_measures': {
-            'perc_windows_tremor': perc_windows_tremor,
-            'median_tremor_power': aggregated_tremor_power['median_tremor_power'],
-            'modal_tremor_power': aggregated_tremor_power['modal_tremor_power'],
-            '90p_tremor_power': aggregated_tremor_power['90p_tremor_power']
+    if perc_windows_tremor == 0: # if no tremor is detected 
+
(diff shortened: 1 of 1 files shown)


src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -183,36 +183,53 @@ def aggregate_tremor(df: pd.DataFrame, config: TremorConfig):
     # calculate tremor time
     perc_windows_tremor= np.sum(df_filtered['pred_tremor_checked']) / nr_windows_rest * 100 # as percentage of total measured time without non-tremor arm movement
```
Erikpostt: Could `nr_windows_rest` be equal to 0? I would perhaps raise an error / warning if this happens (and could happend).

src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -183,36 +183,53 @@ def aggregate_tremor(df: pd.DataFrame, config: TremorConfig):
...
 
-    # calculate aggregated tremor power measures
-    tremor_power = df_filtered.loc[df_filtered['pred_tremor_checked'] == 1, 'tremor_power']
-    tremor_power = np.log10(tremor_power+1) # convert to log scale
-    aggregated_tremor_power = {}
+    if perc_windows_tremor == 0: # if no tremor is detected
```
Erikpostt: I'm not sure, but do you think you'd rather check `perc_windows_tremor` here or `np.sum(df_filtered['pred_tremor_check'])`? If the latter, you can create a variable `n_windows_tremor` or something before line 184.

src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -183,36 +183,53 @@ def aggregate_tremor(df: pd.DataFrame, config: TremorConfig):
...
+                'nr_windows_rest': nr_windows_rest
+            },
+            'aggregated_tremor_measures': {
+                'perc_windows_tremor': perc_windows_tremor,
+                'median_tremor_power': np.nan,
+                'modal_tremor_power': np.nan,
+                '90p_tremor_power': np.nan
+            }
+        }
+
```
Erikpostt: Instead of creating a new dictionary, you could simply create the variable `median_tremor_power` and set it to `np.nan` here, and to `aggregated_tremor_power['median_tremor_power']` down below. Then, outside the if-else statement, create the dictionary `d_aggregates`.

src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -183,36 +183,53 @@ def aggregate_tremor(df: pd.DataFrame, config: TremorConfig):
...
-            aggregated_tremor_power[aggregate_name] = aggregate_parameter(tremor_power, aggregate)
+        for aggregate in config.aggregates_tremor_power:
+            aggregate_name = f"{aggregate}_tremor_power"
+            if aggregate == 'mode':
+                # calculate modal tremor power
+                bin_edges = np.linspace(0, 6, 301)
+                kde = gaussian_kde(tremor_power)
+                kde_values = kde(bin_edges)
+                max_index = np.argmax(kde_values)
+                aggregated_tremor_power['modal_tremor_power'] = bin_edges[max_index]
```
Erikpostt, line 212: Do you envision this being integrated into `aggregate_parameter()`? Like, perhaps the `'aggregate'`  parameter is equal to `'kde_mode'` or something like that? Just curious about your perspective here.

Erikpostt (approved):
Looking good, ready to merge. I've made several suggestions, which you could add before merging this file.

src/paradigma/pipelines/tremor_pipeline.py
nienketimmermans: Yes, good point, this could indeed happen

src/paradigma/pipelines/tremor_pipeline.py
nienketimmermans: Good idea

src/paradigma/pipelines/tremor_pipeline.py
nienketimmermans: It would be nice to have it integrated into `aggregate_paramaeter()`, but the bin sizes that I use for the kernel density estimation are tremor specific, so I'm not sure how to make it also suitable for other pipelines (if that's what we want)

Erikpostt: It indeed looks a little specific, but you could make a separate function that takes as input parameters the power, as well as the start, end, and number of samples for the linear spacing. It works well now, so nothing to worry about currently.

PR #148, opened 2025-02-21
Update pyproject.toml

KarsVeldkamp:


KarsVeldkamp:
@Erikpostt I think we tried this before and it didn't work but I am just curious if this solves it

Issue #169, opened 2025-03-10
Add elaboration on weekly results to each tutorial


Erikpostt:
After #153, the weekly aggregates should be explained to a user now that it is no longer based on 10 minutes of data but actually on weekly data.

Issue #162, opened 2025-02-27
Add datetime subsetting to tutorials


Erikpostt:
As a user, I want to select a range of hours where I want to conduct the aggregation step.

- Let user add start datetime of the dataframe
- Let user select range of days / hours
- Subset dataframe based on selection
- Report aggregation

Issue #153, opened 2025-02-25
Make available one week of PPP data


Erikpostt:
As a user, I want to run the tutorials and return informative weekly parameters. This data could be published on GitHub, or elsewhere.

* Select a subject and week to share
* Select a location to host the data
* Load the data in the tutorials

Erikpostt:
* 1-2 days of data
* At minimum 2 segments
* Not too much data (users need to load data when using package)

PR #179, opened 2025-03-18
Add 2 days of data and subsetting in aggregation

nienketimmermans:
## Description

Solves #153 and partly solves #162 and #169. 

## Changes

- I have added 2 days of data, consisting of 2 segments. 
- I updated the tremor tutorial to process these 2 segments and created the function `select_hours` and `select_days` for subsetting the dataframe for aggregation. 

@Erikpostt let me know if you have any other suggestions. I will update the ppg and gait tutorials afterwards.


docs/tutorials/tremor_analysis.ipynb
```diff
@@ -11,7 +11,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This tutorial shows how to run the tremor pipeline on prepared IMU or gyroscope data to obtain aggregated tremor measures."
+    "This tutorial shows how to run the tremor pipeline to obtain aggregated tremor measures from gyroscope sensor data. Before following along, make sure all data preparation steps have been followed in the data preparation tutorial. In this tutorial, we use two days of data from a participant of the Personalized Parkinson Project to demonstrate the functionalities. The data was stored in two contiguous segments. Per segment we load the data, preprocess the data, extract tremor features and detect tremor. We then combine the output of the different segments to compute aggregated tremor measures."
```
Erikpostt:
```suggestion
    "This tutorial shows how to run the tremor pipeline to obtain aggregated tremor measures from gyroscope sensor data. Before following along, make sure all data preparation steps have been followed in the data preparation tutorial. 
    
    In this tutorial, we use two days of data from a participant of the Personalized Parkinson Project to demonstrate the functionalities. Since _ParaDigMa` expects contiguous time series, the collected data was stored in two segments each with contiguous timestamps. Per segment, we load and preprocess the time series data, extract tremor features and detect tremor. We then combine the output of the different segments to compute aggregated tremor measures."
```

Erikpostt: I tried to emphasize the requirement of contiguous segments and increase readability. Feel free to adjust where needed.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -25,7 +25,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Example IMU data (8 minutes) from a participant of the Personalized Parkinson Project is loaded. The data was prepared as explained in the data preparation tutorial. The prepared data contains both accelerometer and gyroscope data, but only gyroscope data is necessary for running the tremor pipeline."
+    "Load the prepared data into memory. For example, the following functions can be used depending on the file extension of the data:\n",
```
Erikpostt:
```suggestion
    "We use the interally developed `TSDF` ([documentation](https://biomarkersparkinson.github.io/tsdf/)) to load and store data [[1](https://arxiv.org/abs/2211.11294)]. Depending on the file extension of your time series data, examples of other Python functions for loading the data into memory include:\n
    - _.csv_: `pandas.read_csv()` ([documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html))\n",
    "- _.json_: `json.load()` ([documentation](https://docs.python.org/3/library/json.html#json.load))",
```

Erikpostt: Here I removed the "Load the prepared data into memory" because that seemed evident to me given the header "Load example data". Also, I moved the TSDF part up.

Perhaps in the future we could add a TSDF load/store tutorial somewhere so it is properly introduced.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -221,9 +777,11 @@
     "# Set the path to where the prepared data is saved and load the data.\n",
     "# Note: the test data is stored in TSDF, but you can load your data in your own way\n",
     "path_to_data =  Path('../../tests/data')\n",
-    "path_to_prepared_data = path_to_data / '1.prepared_data' / 'imu'\n",
+    "path_to_prepared_data = path_to_data / 'prepared_data_2days' / 'imu'\n",
     "\n",
-    "df_data, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data, prefix='IMU')\n",
+    "segment_nr  = '0001' # Start with the first segment to run steps 1 - 3. By running the cell below step 3 you repeat these steps for the second (and possibly other) segments.\n",
```
Erikpostt, line 782:
```suggestion
    "segment_nr  = '0001' # Start with the first segment to run steps 1 - 3. 
    By running the cell below step 3 you repeat these steps for the second (and possibly other) segments.\n",
```

Erikpostt:
Why did you opt for storing the data in the _tests_ directory? I would personally add a _data_ directory somewhere such that it can be accessed by users. The _tests_ directory is primarily for testing the functionality and robustness of the code.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -35,6 +39,558 @@
    "outputs": [
     {
      "data": {
+      "application/vnd.microsoft.datawrangler.viewer.v0+json": {
```
Erikpostt: Do you have any idea what this output is? I don't see it in the notebook, while it is a lot of lines in the backend. This significantly inreases latency on my side when reviewing.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -256,6 +814,393 @@
     },
     {
      "data": {
+      "application/vnd.microsoft.datawrangler.viewer.v0+json": {
```
Erikpostt: This seems to be reoccuring for each dataframe print.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -221,9 +777,11 @@
     "# Set the path to where the prepared data is saved and load the data.\n",
     "# Note: the test data is stored in TSDF, but you can load your data in your own way\n",
     "path_to_data =  Path('../../tests/data')\n",
-    "path_to_prepared_data = path_to_data / '1.prepared_data' / 'imu'\n",
+    "path_to_prepared_data = path_to_data / 'prepared_data_2days' / 'imu'\n",
     "\n",
-    "df_data, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data, prefix='IMU')\n",
+    "segment_nr  = '0001' # Start with the first segment to run steps 1 - 3. By running the cell below step 3 you repeat these steps for the second (and possibly other) segments.\n",
+    "\n",
+    "df_data, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data, prefix='IMU_segment'+segment_nr)\n",
```
Erikpostt: - Perhaps you could introduce the steps conducted in this tutorial in the introduction, such that it is clear what steps 1 - 3 are when the reader reads this;
- Then, in this section, you could explain that the example here shows how it works for a single contiguous time series, and further down below how it works when someone has multiple "segments" or "contiguous time series";
- I don't think you use `metadata_values` or `metadata_time` anywhere, so you could replace them with `_`. This ensures the user is not confused, that they don't think they need a metadata object;
- I think for readability it would be useful to move the comment (#) above the code line;
- You could also add a link in the markdown cell that, when clicked on, moves the cursor to the cell below steps 1 - 3. In that way, a user does not need to scroll and look for where the cell after step 3 is.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -937,6 +3371,46 @@
     "df_predictions[['time','pred_tremor_proba','pred_tremor_logreg','pred_arm_at_rest','pred_tremor_checked']]"
```
Erikpostt:
```suggestion
    "df_predictions[['time' ,'pred_tremor_proba', 'pred_tremor_logreg', 'pred_arm_at_rest', 'pred_tremor_checked']]"
```

Erikpostt: Also: the line "print(f'A threshold ..." is a little long. It would be best if a user does not need to use the horizontal scroll bar to read the code. Therefore, you could opt for creating new lines every now and again.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -937,6 +3371,46 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Store the predictions of the first segment in a dictionary \n",
+    "dict_predictions = {}\n",
+    "dict_predictions['segment'+segment_nr] = df_predictions\n",
```
Erikpostt: It's up to you, but generally f-strings are common practice for these cases: `dict_predictions[f'segment{segment_nr}']` or `dict_predictions[f'segment_{segment_nr}']`.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -937,6 +3371,46 @@
...
+    "\n",
+    "segments  = ['0002'] # list with all other available segments\n",
```
Erikpostt, line 992: This indeed works with our TSDF data, but it works differently of course for other file types and names. Perhaps you could explain in the markdown cell of _Repeat steps 1 - 3 for the other segment(s)_ that users can modify the for-loop such that it runs through their respective files. For example, users can change `segments` to a list of the filenames of their respective segmented data.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -937,6 +3371,46 @@
...
+    "\n",
+    "for segment_nr in segments:\n",
+    "    \n",
+    "    # Load the data\n",
+    "    df_data, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data, prefix='IMU_segment'+segment_nr)\n",
```
Erikpostt: Similar to a previous comment, I'm not sure whether `metadata_time` and `metadata_values` should be loaded into memory here.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -937,6 +3371,46 @@
...
+    "\n",
+    "    # 1: Preprocess the data\n",
+    "    config = IMUConfig()\n",
+    "    df_preprocessed_data = preprocess_imu_data(df_data, config, sensor='gyroscope', watch_side='left')\n",
```
Erikpostt, line 1001: I know this does not affect your pipeline, but it might still be useful to clarify that a watch side should be provided here.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -937,6 +3371,46 @@
...
+    "\n",
+    "    # 2: Extract features\n",
+    "    config = TremorConfig(step='features')\n",
+    "    df_features = extract_tremor_features(df_preprocessed_data, config)\n",
+    "\n",
+    "    # 3: Detect tremor\n",
+    "    df_predictions = detect_tremor(df_features, config, full_path_to_classifier_package)\n",
+    "    df_predictions[['time','pred_tremor_proba','pred_tremor_logreg','pred_arm_at_rest','pred_tremor_checked']]\n",
+    "    \n",
+    "    # Store the predictions of the current segment in a dictionary \n",
```
Erikpostt: See previous comment about f-strings.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -937,6 +3371,46 @@
...
+    "    config = IMUConfig()\n",
+    "    df_preprocessed_data = preprocess_imu_data(df_data, config, sensor='gyroscope', watch_side='left')\n",
+    "\n",
+    "    # 2: Extract features\n",
+    "    config = TremorConfig(step='features')\n",
+    "    df_features = extract_tremor_features(df_preprocessed_data, config)\n",
+    "\n",
+    "    # 3: Detect tremor\n",
+    "    df_predictions = detect_tremor(df_features, config, full_path_to_classifier_package)\n",
+    "    df_predictions[['time','pred_tremor_proba','pred_tremor_logreg','pred_arm_at_rest','pred_tremor_checked']]\n",
```
Erikpostt: See previous suggestion for spacing.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -948,34 +3422,65 @@
...
+    "The final step is to compute the amount of tremor time and tremor power with the function [`aggregate_tremor`](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/pipelines/tremor_pipeline.py#:~:text=aggregate_tremor), which aggregates over all windows in the input dataframe. Before doing this, we select the hours and days that we want to include in this analysis. In this case we use data collected between 8 am and 10 pm (specified as `config.selected_hours_start` and `config.selected_hours_end`), and days with at least 10 hours of data (`config.min_hours_per_day`). Based on the selected data, we compute aggregated measures for tremor time and tremor power:\n",
```
Erikpostt: A couple of questions and suggestions about this cell:
- Perhaps you could clarify that datetime selection is optional and depends on the availability of datetime data;
- The `metadata_time` and `metadata_values` were loaded before -- perhaps you could keep them in memory by storing them in a list? Alternatively, you could simply store the `metadata.start_iso8601` when loading into memory somewhere previously in the tutorial when loading the prepared data.
- You could also opt for adding `segment_nr` as a column to `df_predictions` in the cell "Repeat steps 1 - 3 ..." instead of creating a dictionary of dataframes. In that way, you don't need to loop through the segments here again, but simply subset by using `select_hours`. 
- I would declare `config = TremorConfig()` not inside the function call but before.
- Concatenating dataframes in a for-loop is slower than appending dataframes to a list and concatenating them after the loop finished.
- For the for-loop, you can also assign two variables to `dict_predictions.items()` instead of `dict_predictions.keys()`, such that you get: `for segment, predictions_segments in dict_predictions.items()`. Then you don't need to declare `df_predictions_segment = dict_predictions[segment]` afterwards
- Could you perhaps add the reference for why this subsetting is done?
- The final lines outside the loop also declare variables inside the function call: I would put them before the function call.
- Perhaps you could print for each segment the number of valid hours? Such that the user understands why only one of the segments remained.
- It might also be useful to explicitly set `config.selected_hours_start = "08:00"` for example here such that a user can change it if desired. Also, can a user set minutes? That is, "08:23" for example? This also might be useful to clarify (also specify that it is in "HH:MM" format for example)

src/paradigma/config.py
```diff
@@ -223,6 +223,9 @@ def __init__(self, step: str | None = None) -> None:
         # Aggregation
         # -----------
         self.aggregates_tremor_power: List[str] = ['mode', 'median', '90p']
+        self.selected_hours_start: str = '08:00'
```
Erikpostt: It would be useful to specify that this is in "HH:MM" format here

src/paradigma/util.py
```diff
@@ -432,3 +430,65 @@ def merge_predictions_with_timestamps(
     df_ts = df_ts.dropna(subset=[pred_proba_colname])
 
     return df_ts
+
+
+def select_hours(df: pd.DataFrame, start_time: datetime, config) -> pd.DataFrame:
```
Erikpostt: For such a util function, wouldn't it be more useful to simply have `selected_hours_start` and `selected_hours_end` as parameters instead of having to include a config file?

src/paradigma/util.py
```diff
@@ -432,3 +430,65 @@ def merge_predictions_with_timestamps(
...
+        - selected_hours_end: str
+            The end time of the selected hours.
+
+    Returns
+    -------
+    pd.DataFrame
+        The selected data.
+
+    """
+    df = df.assign(datetime_col=start_time + (df['time'].values * timedelta(seconds=1))) # create datetime column
```
Erikpostt: Can't this be done more simply like this?

`df['datetime_col'] = start_time + pd.to_timedelta(df['time'], unit="s")`

I don't see what the `timedelta` is doing besides changing it to seconds perhaps? I think this `pandas` functionality is more efficient and easier to interpret.

src/paradigma/util.py
```diff
@@ -432,3 +430,65 @@ def merge_predictions_with_timestamps(
...
+    Select days of interest from the data to include in the aggregation step.
+
+    Parameters
+    ----------
+    df : pd.DataFrame
+        Input data with column 'datetime_col'.
+
+    config: object
+        Configuration object. The following attribute is used:
+        - min_hours_per_day: int
```
Erikpostt: And window length ;)

src/paradigma/util.py
```diff
@@ -432,3 +430,65 @@ def merge_predictions_with_timestamps(
...
+    Returns
+    -------
+    pd.DataFrame
+        The selected data.
+
+    """
+                
+    df['date'] = df['datetime_col'].dt.date
+    min_windows_per_day = config.min_hours_per_day*3600/config.window_length_s 
+    #df_subset = df.groupby('date').filter(lambda x: len(x) >= min_windows_per_day).drop(columns=['date'])
```
Erikpostt: Should this line be removed?

src/paradigma/util.py
Erikpostt: Also, I don't think you want the '_col' in the column name do you? Perhaps you could change it to 'time_dt' or something similar (such that it links to the 'time' column).

src/paradigma/util.py
```diff
@@ -432,3 +430,65 @@ def merge_predictions_with_timestamps(
...
+            The end time of the selected hours.
+
+    Returns
+    -------
+    pd.DataFrame
+        The selected data.
+
+    """
+    df = df.assign(datetime_col=start_time + (df['time'].values * timedelta(seconds=1))) # create datetime column
+    df_subset = df.set_index('datetime_col').between_time(config.selected_hours_start, config.selected_hours_end).reset_index() # select the hours of interest
```
Erikpostt:
```suggestion
    df_subset = df[df['datetime_col'].dt.time.between(config.selected_hours_start, config.selected_hours_end)]# select the hours of interest
```

Erikpostt: Then you don't have to set the index and reset it again.

src/paradigma/util.py
```diff
@@ -432,3 +430,65 @@ def merge_predictions_with_timestamps(
...
+
+    Returns
+    -------
+    pd.DataFrame
+        The selected data.
+
+    """
+                
+    df['date'] = df['datetime_col'].dt.date
+    min_windows_per_day = config.min_hours_per_day*3600/config.window_length_s
```
Erikpostt:
```suggestion
    min_windows_per_day = config.min_hours_per_day * 3600 / config.window_length_s 
```

src/paradigma/util.py
```diff
@@ -432,3 +430,65 @@ def merge_predictions_with_timestamps(
...
+            The minimum number of hours per day required for including the day in the aggregation step.
+
+    Returns
+    -------
+    pd.DataFrame
+        The selected data.
+
+    """
+                
+    df['date'] = df['datetime_col'].dt.date
```
Erikpostt: I'm not sure whether you want to create more columns here (note that the dataframes can become quite large this way). You could change the final line in this function to:

`df_subset = df.groupby(df['datetime_col'].dt.date).filter(lambda x: len(x) >= min_windows_per_day)`

Then you don't need new columns. Unless you want them explicitly of course.

Erikpostt (changes requested):
Good work! I have quite a few comments and questions, but I think we'll be able to work them out.

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Yes, I used this for another project, it should be removed now.

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Good point, for me it's visible, but maybe my screen is wider than usual ;)

src/paradigma/util.py
nienketimmermans: Yes, that is also an option, but shouldn't we do the same for the selection of ours per day then?

src/paradigma/util.py
nienketimmermans: Yes, forgot to do that

src/paradigma/util.py
nienketimmermans: Good point, I do use the date in the aggregate function to extract the number of unique days and store this as metadata, but that could also be based on the datetime values of course.

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: If we concatenate the dataframes, I think we should replace the time column with relative seconds since segment start to datetime values, otherwise this might be a little confusing. What do you think @Erikpostt, is this still a better way than creating a dictionary?

Erikpostt: Hmm in that case the dictionary might be the most suitable. I noticed in step 4 that you use concatenation, which possibly makes sense there because you want to use `select_days` on the segments concatenated. You could, however, also opt for using a similar approach here by using a dictionary, and then subset each dataframe using `select_days`. I'll leave that to you. Otherwise, it would be useful to clarify why concatenation is used in step 4 and a dictionary in steps 1-3.

src/paradigma/util.py
Erikpostt: Could you clarify? Which block of code are you referring to here?

nienketimmermans: In the select_days function I also use the config file, both for number of hours per day and the window length. So I thought it would make sense to use the same approach in both functions

Erikpostt: I see your point. I think it would be best to omit config objects from the util functions, and only include them in the pipeline functions. You could simply add the hyperparameter indicating the duration (in seconds) of a single row. Then, when calling the function, you can assign the config object as follows:

`min_hours_per_day = 8`
`min_s_per_day = min_hours_per_day * 3600`
`some_function(interval_s = min_s_per_day  / config.window_length_s)`

Or something along those lines?

nienketimmermans: I don't completely understand what you mean here, because you're still using the config object here? I can indeed derive the window length in seconds from the dataframe and then omit the config objects in these functions

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: I concatenated the dataframes in step 4 after adding the column `time_dt`, taking into account the start time of each of the segments. I can also do that earlier, then concatenating makes sence for both steps 1-3 and step 4.

nienketimmermans: About two other points that you mentioned: 

- Could you perhaps add the reference for why this subsetting is done?
_I think it's better to refer to the papers than elaborate here on all the choices that were made_

- Perhaps you could print for each segment the number of valid hours? Such that the user understands why only one of the segments remained.
_The number of valid hours is not specificied per segment, but per day. I did save the number of valid days, but in theory one day could be composed of multiple segments_

Erikpostt: Got it, and I agree. So it would be possible to print the number of valid hours per day at the end? I'll leave it up to you whether you consider it added value.

Erikpostt (approved):
Should be ready to merge. Personally I'm not a big fan of how we use TSDF currently, I would rather specify the filenames in a list that should be loaded using a TSDF function, since this is more similar to loading methods of other data formats. Let's leave that for a later discussion on TSDF.

PR #180, opened 2025-03-25
update_tutorials

nienketimmermans:
## Description
Solves #169 

## Changes
I updated the gait and heart rate analysis tutorials that now process 2 segments of raw data as well.


docs/tutorials/gait_analysis.ipynb
```diff
@@ -1461,58 +1840,191 @@
     "print(f\"Gait segments are created of minimum {config.min_segment_length_s} seconds and maximum {config.max_segment_gap_s} seconds gap between segments.\\n\")\n",
     "print(f\"A total of {quantified_arm_swing['segment_nr'].nunique()} {dataset_used} gait segments have been quantified.\")\n",
     "\n",
-    "print(f\"\\nSegment metadata:\")\n",
-    "pprint(segment_meta)\n",
+    "print(f\"\\nGait segment metadata:\")\n",
+    "pprint(gait_segment_meta)\n",
```
Erikpostt: I would suggest only showing the quantification aggregates across segments and the first segment, since the `quantified_arm_swing` is also only shown for the first segment.

```suggestion
    "pprint(gait_segment_meta['aggregated'], gait_segment_meta['per_segment'][1])\n",
```

docs/tutorials/gait_analysis.ipynb
```diff
@@ -1461,58 +1840,191 @@
...
+    "import pandas as pd\n",
+    "\n",
+    "# Create a list to store all quantified arm swing segments \n",
+    "list_quantified_arm_swing = []\n",
+    "\n",
+    "# Store the quantified arm swing segments of the first raw data segment in the list\n",
+    "quantified_arm_swing['raw_data_segment_nr'] = raw_data_segment_nr\n",
+    "list_quantified_arm_swing.append(quantified_arm_swing)\n",
+    "\n",
+    "raw_data_segments  = ['0002'] # list with all other available raw data segments\n",
```
Erikpostt: If it doesn't take too long to process, I would include both segments '0001' and '0002' here. In that way, this section is an alternative to the sections above, and not necessarily a compliment. This does require the modules to be imported in this cell too though. What do you think about this idea?

docs/tutorials/gait_analysis.ipynb
```diff
@@ -1461,58 +1840,191 @@
...
+    "        raise ValueError(\"No gait detected in the input data.\")\n",
+    "    gait_preprocessing_config = GaitConfig(step='gait')\n",
+    "    \n",
+    "    df = merge_predictions_with_timestamps(\n",
+    "        df_ts=df_preprocessed, \n",
+    "        df_predictions=df_gait, \n",
+    "        pred_proba_colname=DataColumns.PRED_GAIT_PROBA,\n",
+    "        window_length_s=gait_preprocessing_config.window_length_s,\n",
+    "        fs=gait_preprocessing_config.sampling_frequency\n",
+    "    )\n",
```
Erikpostt: I don't think `gait_preprocessing_config` needs to be defined here, as `config` is an identical object.

```suggestion
    "    df = merge_predictions_with_timestamps(\n",
    "        df_ts=df_preprocessed, \n",
    "        df_predictions=df_gait, \n",
    "        pred_proba_colname=DataColumns.PRED_GAIT_PROBA,\n",
    "        window_length_s=config.window_length_s,\n",
    "        fs=config.sampling_frequency\n",
    "    )\n",
```

docs/tutorials/gait_analysis.ipynb
```diff
@@ -1461,58 +1840,191 @@
...
+    "    df_arm_activity['pred_no_other_arm_activity_proba'] = filter_gait(\n",
+    "        df=df_arm_activity,\n",
+    "        clf_package=clf_package_filtering\n",
+    "    )\n",
+    "\n",
+    "    # Merge arm activity predictions into timeseries data\n",
+    "    if not any(df_arm_activity[DataColumns.PRED_NO_OTHER_ARM_ACTIVITY_PROBA] >= clf_package_filtering.threshold):\n",
+    "        raise ValueError(\"No gait without other arm activities detected in the input data.\")\n",
+    "\n",
+    "    config = GaitConfig(step='arm_activity')\n",
```
Erikpostt: Same as previous comment.

```suggestion
```

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -280,19 +288,23 @@
...
-    "path_to_prepared_data = path_to_data / '1.prepared_data'\n",
+    "# Set the path to where the prepared data is saved and load the data.\n",
+    "# Note: the test data is stored in TSDF, but you can load your data in your own way\n",
+    "path_to_prepared_data =  Path('../../example_data')\n",
+    "\n",
+    "ppg_prefix = 'ppg'\n",
+    "imu_prefix = 'imu'\n",
+    "\n",
+    "segment_nr  = '0001' \n",
     "\n",
```
Erikpostt:
```suggestion
    "\n",
    "segment_nr = '0001' \n",
```

docs/tutorials/heart_rate_analysis.ipynb
```diff
@@ -1174,6 +1219,101 @@
...
+    "import pandas as pd\n",
+    "\n",
+    "# Create a list of dataframes to store the estimated heart rates of all segments\n",
+    "list_df_hr = []\n",
+    "\n",
+    "# Store the hr estimations of the first segment in the list\n",
+    "df_hr['segment_nr'] = segment_nr\n",
+    "list_df_hr.append(df_hr)\n",
+    "\n",
+    "segments  = ['0002'] # list with all other available segments\n",
```
Erikpostt: Same suggestion as for gait. Also:

```suggestion
    "segments = ['0002'] # list with all other available segments\n",
```

docs/tutorials/tremor_analysis.ipynb
Erikpostt, line 1: Your execution count is off, but not a reason to not merge.

Erikpostt (approved):
Great work! I have some minor suggestions.

docs/tutorials/gait_analysis.ipynb
nienketimmermans: Yes, that's a good idea, it will also become easier for users to apply it to their own data then. I will also do this for the other tutorials

nienketimmermans: Althoug for the heart rate pipeline it took 2m15s to repeat the steps for the second segment. Do you think that's too long?

Erikpostt: I think that's fine, it's a little long but the user can adjust it (and the user should knows this if they process one segment first)

Issue #163, opened 2025-02-27
Load/store data as TSDF in tutorials


Erikpostt:
Double-check whether we use TSDF consistently in the tutorials: load once (raw data), and store once (classification & quantification). A line of code / cell block should be provided to store the TSDF data in the final step (not aggregation, this is not stored / can easily be stored by a user as .csv or something).

PR #182, opened 2025-04-07
Store quantification as tsdf

Erikpostt:
## Description

See #163. The tutorials using example data only used TSDF for loading data, but not for storing. 

## Changes

I have added the option to store a time series dataframe in the domain tutorials: (1) gait, (2) tremor, and (3) heart rate. These have been added after the classification step (for gait after gait detection, such that a user can move it to detection of other arm activities if desired).

## Additional Information

We had to make a decision whether to store the classification and/or quantification step, but the quantification step for gait does not result in a time series. I therefore opted for providing example code for storing in TSDF format at the classification step.

## Checklist

- [X] Tests passed
- [X] Documentation updated

### DIFF
--- a/docs/tutorials/gait_analysis.ipynb
+++ b/docs/tutorials/gait_analysis.ipynb
@@ -239,7 +239,7 @@
     "raw_data_segment_nr  = '0001' \n",
     "\n",
     "# Load the data from the file\n",
-    "df_imu, _, _ = load_tsdf_dataframe(path_to_prepared_data, prefix=f'IMU_segment{raw_data_segment_nr}')\n",
+    "df_imu, metadata_time, metadata_values = load_tsdf_dataframe(path_to_prepared_data, prefix=f'IMU_segment{raw_data_segment_nr}')\n",
     "\n",
     "df_imu"
    ]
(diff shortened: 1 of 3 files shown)


docs/tutorials/gait_analysis.ipynb
```diff
@@ -832,6 +832,127 @@
     "df_gait[['time', 'pred_gait_proba']].head()"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Store as TSDF\n",
+    "The predicted probabilities (and optionally other features) can be stored and loaded in TSDF as demonstrated below. "
```
nienketimmermans, line 840: You only show how to store the data here, so maybe remove _and loaded_?

nienketimmermans (approved):
Makes sense to only store the classification output in TSDF. I have a small comment, but other than that it's ready to be merged.

docs/tutorials/gait_analysis.ipynb
Erikpostt: The cell beneath this cell actually loads the data. Do you still think we should remove it?

nienketimmermans: Ah yes I overlooked that, then you can keep it like this. Would indeed be helpful to also show how you can load it.

PR #186, opened 2025-05-12
Remove capitalization and rename gait to arm swing during gait

Erikpostt:
## Description

The main reason is to change the 'Gait' domain to 'Arm swing during gait' to align with the naming convention used in the Paradigma paper. I also changed the capitalization of all words which I think was unnecessary.

## Changes

* Change column 'Gait' to 'Arm swing during gait'
* Change capitalized words to lower


Erikpostt:
Set to draft, changes should be discussed in upcoming meeting 19/05.

nienketimmermans (changes requested):
During the meeting on 19/05 we agreed to change _probability_ to _predictions_ for the tremor and gait pipelines, since this is the main output of the predictions step.

nienketimmermans (approved):

KarsVeldkamp (approved):

Issue #187, opened 2025-05-19
Tremor pipeline: align with other domains


Erikpostt:
The classification -> quantification -> aggregation steps of the tremor pipeline are currently inconsistent with the other pipelines, and not in line with the diagram in the readme.

PR #189, opened 2025-05-22
Align tremor pipeline

nienketimmermans:
## Description

Solves: #187

## Changes

Adds the quantification step to the tremor pipeline, so that it aligns with the other pipelines.

## Additional Information

To make sure that the pipeline remains efficient, the tremor power is already calculated in the feature extraction step of the pipeline, and then selected for the predicted tremor windows in the quantification step.

## Checklist

- [x] Tests passed
- [x] Documentation updated


Erikpostt (approved):
Checked, ready to merge. You could change the order of columns such that `time_dt` column is next to `time` for clarity, but otherwise feel free to merge.

nienketimmermans:
Good idea, I have changed the order of columns now.

PR #206, opened 2025-09-02
Add flexibility segment categories

Erikpostt:
## Description

This PR contains changes to the gait pipeline primarily, but also to a minor extend generally. For the gait pipeline, I wanted to add more flexibility to the user by removing pre-determined gait segment categories (and durations). 

## Changes
To the gait pipeline:
* Remove pre-defined gait segment categories;
* Add aggregation per gait segment to the arm swing aggregation step, which was previously only done at the week level;
* Clarify distinction between unfiltered and filtered gait;

And, more generally:
* Increase maximum threshold for contiguous timestamps, since quite a few timestamps were ~ 2 * 1 / fs apart. 
* Add coefficient of variation (CoV) to list of aggregations;
* Restructure and document .gitignore


docs/tutorials/gait_analysis.ipynb
```diff
@@ -43,7 +43,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 16,
```
nienketimmermans: Do you want to change these execution counts?

docs/tutorials/pulse_rate_analysis.ipynb
```diff
@@ -4285,16 +1467,16 @@
...
    "metadata": {},
    "outputs": [
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "{'metadata': {'nr_pr_est': 806},\n",
-      " 'pr_aggregates': {'99p_pulse_rate': 87.65865011636926,\n",
-      "                   'mode_pulse_rate': 81.25613346418058}}\n"
+      "{'metadata': {'nr_pr_est': 7854},\n",
```
nienketimmermans: Why did this change?

nienketimmermans (changes requested):
@Erikpostt most of the changes are looking good! Maybe you can clarify in step 7 of the gait_analysis notebook that the user can define the segment categories, now it's only shown in the code itself. 
Another remark is that the pulse rate estimates have changed. Was that on purpose?

docs/tutorials/pulse_rate_analysis.ipynb
Erikpostt: Since the previous version did not run multiple segments (penultimate cell), while I ran all cells. It should be fixed now.

Erikpostt:
@nienketimmermans thanks for reviewing, I've implemented the necessary changes. Can you take a look?

nienketimmermans (approved):

PR #207, opened 2025-09-02
Add resampling frequency 

nienketimmermans:
## Description

This PR solves two bugs: 1) It was not possible to resample to another frequency than the sampling frequency, and 2) there was a bug in the `select_days` function that was used in the tremor analysis (the array 'window_length_s' was assumed to start at index 0, but this is not always the case).

## Changes

1. I added the variable resampling frequency in the config file and changed the function `preprocess_imu_data` accordingly.
2. The length of the window in the `select_days` function is now determined based on the correct index.

## Checklist

- [x] Tests passed
- [x] Documentation updated


Erikpostt (commented):
Good spot. About the second change; why do you index [1] (i.e., the second row) instead of [0] (i.e., the first row)?

nienketimmermans:
Because the diff function returns NaN as the first row.

Erikpostt (approved):
Ready to merge.

Issue #164, opened 2025-02-27
Add modal tremor power to feature statistics function


Erikpostt:
The modal calculated for tremor power does not suit the statistics in `compute_statistics` in [feature_extraction](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/feature_extraction.py). We should try to rewrite the modal calculation of the tremor power and integrate it in the `compute_statistics` function.

PR #212, opened 2025-09-05
 Add modal tremor power to feature statistics function

nienketimmermans:
## Description

Solves #164 

## Changes

The bin edges used for computing the modal tremor power are now an optional parameter in the `aggregate_parameter` function. 

## Checklist

- [x] Tests passed
- [x] Documentation updated


nienketimmermans:
Only the last commit is relevant

docs/tutorials/tremor_analysis.ipynb
Erikpostt, line 1: Any particular reason why you're no longer displaying this output?

src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -200,17 +199,10 @@ def aggregate_tremor(df: pd.DataFrame, config: TremorConfig):
         tremor_power = df_filtered.loc[df_filtered['pred_tremor_checked'] == 1, 'tremor_power']
         tremor_power = np.log10(tremor_power+1) # convert to log scale
         
+        bin_edges = np.linspace(0, 6, 301) # bin edges for calculating the mode of tremor power in log scale
```
Erikpostt: I think you'd want to have this somewhere in your Config object, in `config.py`?

src/paradigma/util.py
```diff
@@ -337,6 +342,11 @@ def aggregate_parameter(parameter: np.ndarray, aggregate: str) -> np.ndarray | i
         return np.mean(parameter)
     elif aggregate == 'median':
         return np.median(parameter)
+    elif aggregate == 'mode' and bin_edges is not None:
```
Erikpostt: If aggregate equals `mode` and bin_edges equals `None` you probably want to raise an error that bin_edges should be defined when using mode. Now, this if-statement will pass and end up at `raise ValueError(f"Invalid aggregation method: {aggregate}")`, which is not exactly correct I think.

Erikpostt (changes requested):
I've left some suggestions, feel free to take a look and implement the changes if need be.

src/paradigma/pipelines/tremor_pipeline.py
nienketimmermans: Yes, good idea

src/paradigma/util.py
nienketimmermans: Good point. If bin_edges equals `None`, the function will now continue using the other function for computing the mode (which is used for modal pulse rate), which will not return correct values for tremor power. To make this distinction more clear, I changed the aggregate 'mode' to 'mode_binned' for tremor power. If no bin_edges are provided, it will raise an error that these should be defined.

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: This cell just stores the output in TSDF format

nienketimmermans:
@Erikpostt thank you for reviewing, can you have another look?

Erikpostt (approved):
Looks solid now, just one more proposal: in the tremor tutorial you clarify at step 5 that you compute the median, 90th percentile and the mode. Perhaps you'd want to explain briefly how the mode is calculated? Or would you rather refer to the scientific publication / source code? I'll leave it up to you.

Issue #176, opened 2025-03-17
Scale sensitivity


Erikpostt:
@KarsVeldkamp can you create this issue and assign it to the release?

Issue #167, opened 2025-02-27
Make addition of accelerometer optional (PPG)


Erikpostt:
Currently, the inclusion of accelerometer is required for the PPG pipeline. However, this should be made optional.

PR #220, opened 2025-09-23
Ppg scale sensitivity + optional accelerometer

KarsVeldkamp:
## Description

This PR solves #176 and #167 

## Changes
For the PPG scale sensitivity, I updated the tutorial with additional guidance on how users should handle this. In future releases, we will explore options to make the classifier scale-insensitive and/or provide an easy way to obtain scaling parameters and apply them immediately.

## Checklist

- [x] Tests passed
- [x] Documentation updated


docs/tutorials/_static/pulse_rate_analysis.md
Erikpostt, line 1: Make sure it is clear to the user why motion artifacts are an issue when using PPG data, and specifically for daytime analysis. If this requires more elaboration than a quick tutorial allows, you could also make a feature request to write a separate short documentation file on PPG signal + motion artifacts in subsequent release.

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -39,12 +39,14 @@ df_ppg, metadata_time_ppg, metadata_values_ppg = load_tsdf_dataframe(
...
 )
+
+# Only relevant if you have IMU data available
 df_imu, metadata_time_imu, metadata_values_imu = load_tsdf_dataframe(
     path_to_data=path_to_prepared_data / imu_prefix, 
     prefix=f'IMU_segment{segment_nr}'
 )
 
-# Drop the gyroscope columns from the IMU data
+# Drop the gyroscope columns from the IMU data to keep only accelerometer data
```
Erikpostt: Users may only have accelerometer column, or users may have other columns beside gyroscope columns. Wouldn't it be more useful to specifically select a timestamp column and the accelerometer columns?

Erikpostt: Also, the '#' implies a new section after the 'Only relevant if you have IMU data available' section, while the 'Drop the gyroscope columns' this applies to this section. Slightly confusing.

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -477,7 +481,7 @@ display(df_ppg_proc, df_acc_proc)
 
 ## Step 2: Extract signal quality features
 
-The preprocessed data (PPG & accelerometer) is windowed into overlapping windows of length `ppg_config.window_length_s` with a window step of `ppg_config.window_step_length_s`. From the PPG windows 10 time- and frequency domain features are extracted to assess PPG morphology and from the accelerometer windows one relative power feature is calculated to assess periodic motion artifacts.
+The preprocessed data (PPG & accelerometer) is windowed into overlapping windows of length `ppg_config.window_length_s` with a window step of `ppg_config.window_step_length_s`. From the PPG windows 10 time- and frequency domain features are extracted to assess PPG morphology and from the accelerometer windows (optional) one relative power feature is calculated to assess periodic motion artifacts.
```
Erikpostt: Accelerometer in first line is also optional. Perhaps just remove the 'PPG & accelerometer' bit between brackets, and clarify in the final line that in case of accelerometry the relative power feature is calculated. Also, the line "from the PPG..." does not read easily, I'd suggest perhaps adding a comma (,) after "PPG windows".

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -717,10 +722,15 @@ df_features
...
 
 The classification step is implemented in `signal_quality_classification` (documentation can be found [here](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/pipelines/pulse_rate_pipeline.py#:~:text=signal_quality_classification)).
 
+<u>**_Note on scale sensitivity_**</u>  
+The PPG sensor used for developing this pipeline records in arbitrary units. Some features are scale sensitive and require rescaling when applying the pipeline to other datasets or PPG sensors.  
+In this pipeline, the logistic classifier for PPG morphology was trained on z-scored features, using the means (μ) and standard deviations (σ) from the Personalized Parkinson Project training set. These μ and σ values are stored in the `ppg_quality_classifier_package`.
```
Erikpostt: What's the training set? Perhaps you can link to the specific page in your publication.

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -717,10 +722,15 @@ df_features
...
 
 The classification step is implemented in `signal_quality_classification` (documentation can be found [here](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/pipelines/pulse_rate_pipeline.py#:~:text=signal_quality_classification)).
 
+<u>**_Note on scale sensitivity_**</u>  
+The PPG sensor used for developing this pipeline records in arbitrary units. Some features are scale sensitive and require rescaling when applying the pipeline to other datasets or PPG sensors.  
+In this pipeline, the logistic classifier for PPG morphology was trained on z-scored features, using the means (μ) and standard deviations (σ) from the Personalized Parkinson Project training set. These μ and σ values are stored in the `ppg_quality_classifier_package`.  
+When applying the code to another dataset, users are advised to recalculate **_μ_** and **_σ** for each feature on their (training) data and update the classifier package accordingly.
```
Erikpostt: If this is the case, I think it requires a bit more elaboration. For example, how do you aggregate? I guess this is simple output from a Scikit object, but I'm not sure. You might want to refer users to a webpage where it is instructed how to do this (e.g., scikit package module).

Also, how do they update the classifier package?

docs/tutorials/pulse_rate_analysis.ipynb
Erikpostt, line 1: See comments for markdown.

src/paradigma/pipelines/pulse_rate_utils.py
```diff
@@ -8,7 +8,7 @@
 def assign_sqa_label(
         ppg_prob: np.ndarray, 
         config: PulseRateConfig, 
-        acc_label=None
+        acc_label: np.ndarray | None = None
```
Erikpostt: What I'm missing here is what happens if I don't input `acc_label`. The parameter currently states in the docstring:

```
acc_label : np.ndarray, optional
        The labels for the accelerometer.
```

It should have a default value too.

src/paradigma/preprocessing.py
```diff
@@ -245,8 +245,8 @@ def preprocess_imu_data(df: pd.DataFrame, config: IMUConfig, sensor: str, watch_
     return df
 
 
-def preprocess_ppg_data(df_ppg: pd.DataFrame, df_acc: pd.DataFrame, ppg_config: PPGConfig, 
-                        imu_config: IMUConfig, start_time_ppg: str, start_time_imu: str) -> Tuple[pd.DataFrame, pd.DataFrame]:
+def preprocess_ppg_data(df_ppg: pd.DataFrame, ppg_config: PPGConfig, start_time_ppg: str, start_time_imu: str,
```
Erikpostt: It is unclear to me what is preprocessed for ppg data in this function when reading the docstring. Perhaps you'd want to clarify what is done with accelerometer data, and what is done with PPG data separately.

Erikpostt: Additionally, calculating overlapping segments is worth a slightly more extensive introduction. You put in hard effort to make this work and it's clean code.

src/paradigma/preprocessing.py
```diff
@@ -333,8 +335,12 @@ def preprocess_ppg_data(df_ppg: pd.DataFrame, df_acc: pd.DataFrame, ppg_config:
...
         # Replace or add new columns based on configuration
         df_ppg_proc[filter_config["result_columns"]] = filtered_data
-    
-    return df_ppg_proc, df_acc_proc
+
+    if df_acc is not None and imu_config is not None:
+        return df_ppg_proc, df_acc_proc
+    else:
+        df_acc_proc = None
+        return df_ppg_proc, df_acc_proc
```
Erikpostt: The `return` statement is done in both cases, so can be moved outside the loop. You only need to set `df_acc_proc` inside a loop.

Erikpostt (changes requested):
I've made some suggestions that would be useful to implement prior to merging into main.

Extra note: next time, please make sure either the issue is fully descriptive (preferable) or it is clear in the PR why scale sensitivity should be solved, and how this can be solved. This makes it easier to create a changelog at release and ensure users understand the changes made.

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -14,7 +14,7 @@ We then combine the output of the different segments for the final step:
 
 ## Load data
 
-This pipeline requires accelerometer and PPG data to run. Here, we start by loading a single contiguous time series (segment), for which we continue running steps 1-4. [Below](#multiple_segments_cell) we show how to run these steps for multiple segments. The channel `green` represents the values obtained with PPG using green light.
+This pipeline requires PPG data and preferably accelerometer data (optional). Here, we start by loading a single contiguous time series (segment), for which we continue running steps 1-4. [Below](#multiple_segments_cell) we show how to run these steps for multiple segments. The channel `green` represents the values obtained with PPG using green light.
```
Erikpostt: Personally I think this phrasing is slightly confusing, because you start with "requires", but accelerometer data is optional (now it states requires accelerometer data preferably). Je mag het zo laten, maar het kan ook als volgt:
1. PPG data is required for this pipeline; accelerometer data is optional but recommended.
2. This pipeline requires PPG data and can be enhanced with optional accelerometer data.
3. The pipeline requires PPG data and can optionally use accelerometer data.

docs/tutorials/_static/pulse_rate_analysis.md
Erikpostt: You might've missed this comment, but it is still unclear whether "Drop the gyroscope ..." falls under "Only relevant if ..." or not. It should, but it would be useful to clarify in a way.

docs/tutorials/_static/pulse_rate_analysis.md
Erikpostt: Again, I'm not sure whether you fully addressed this issue. If you don't agree, you can mention that, otherwise I'd suggest splitting the final sentence up for readability purposes.

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -746,10 +751,15 @@ df_features
 
 ## Step 3: Signal quality classification
 
-A trained logistic classifier is used to predict PPG signal quality and returns the `pred_sqa_proba`, which is the posterior probability of a PPG window to look like the typical PPG morphology (higher probability indicates toward the typical PPG morphology). The relative power feature from the accelerometer is compared to a threshold for periodic artifacts and therefore `pred_sqa_acc_label` is used to return a label indicating predicted periodic motion artifacts (label 0) or no periodic motion artifacts (label 1).
+A trained logistic classifier is used to predict PPG signal quality and returns the `pred_sqa_proba`, which is the posterior probability of a PPG window to look like the typical PPG morphology (higher probability indicates toward the typical PPG morphology). The (optional) relative power feature from the accelerometer is compared to a threshold for periodic artifacts and therefore `pred_sqa_acc_label` is used to return a label indicating predicted periodic motion artifacts (label 0) or no periodic motion artifacts (label 1).
```
Erikpostt: I would clarify this more in case of accelerometer, for example using a new line and starting with: "If accelerometer data is used, the relative power feature ...".

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -746,10 +751,15 @@ df_features
...
 
 The classification step is implemented in `signal_quality_classification` (documentation can be found [here](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/pipelines/pulse_rate_pipeline.py#:~:text=signal_quality_classification)).
 
+<u>**_Note on scale sensitivity_**</u>
+The PPG sensor used for developing this pipeline records in arbitrary units. Some features are scale sensitive and require rescaling when applying the pipeline to other datasets or PPG sensors.
+In this pipeline, the logistic classifier for PPG morphology was trained on z-scored features, using the means (μ) and standard deviations (σ) from the Personalized Parkinson Project training set. These μ and σ values are stored in the `ppg_quality_classifier_package`.
+When applying the code to another dataset, users are advised to recalculate **_μ_** and **_σ** for each feature on their (training) data and update the classifier package accordingly.
```
Erikpostt: I think you're missing a "_" after sigma.

docs/tutorials/pulse_rate_analysis.ipynb
```diff
@@ -64,14 +64,16 @@
...
+    "# Only relevant if you have IMU data available\n",
     "df_imu, metadata_time_imu, metadata_values_imu = load_tsdf_dataframe(\n",
     "    path_to_data=path_to_prepared_data / imu_prefix, \n",
     "    prefix=f'IMU_segment{segment_nr}'\n",
     ")\n",
     "\n",
-    "# Drop the gyroscope columns from the IMU data\n",
-    "cols_to_drop = df_imu.filter(regex='^gyroscope_').columns\n",
-    "df_acc = df_imu.drop(cols_to_drop, axis=1)\n",
+    "time_col = df_imu.filter(regex='^time$').columns # select time column\n",
```
Erikpostt: Why don't you let users simply declare the time columns? For example:

`time_col = ['time']`.

And let users rewrite according to their preferences. I personally think using `regex` is kind of like killing a fly with a cannon, and may cause unexpected errors (e.g., what happens if another column has the string `time` in it?). For me, the same holds for the accelerometer columns, so instead use `accelerometer_x`, `accelerometer_y`, etc., and let users redefine.

docs/tutorials/pulse_rate_analysis.ipynb
```diff
@@ -457,7 +475,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "paradigma (Poetry)",
```
Erikpostt: I expected the pre-commit hooks to remove any metadata of the notebooks. Do you have any idea why this is still shown?

src/paradigma/classification.py
```diff
@@ -47,6 +48,18 @@ def transform_features(self, X) -> np.ndarray:
...
+        """
+        Update the standard scaler used for feature transformation.
+
+        Parameters
+        ----------
+        x_train : np.ndarray
+            The training data to fit the scaler.
+        """
+        scaler = StandardScaler()
+        self.scaler = scaler.fit(x_train)
```
Erikpostt:
```suggestion
    def update_scaler(self, x_train: np.ndarray) -> None:
        """
        Update the scaler used for feature transformation.

        Parameters
        ----------
        x_train : np.ndarray
            The training data to fit the scaler.
        """
        scaler = StandardScaler()
        self.scaler = scaler.fit(x_train)
```

Erikpostt: I would either use `scaler` or `StandardScaler()` in the docstring.

src/paradigma/pipelines/pulse_rate_pipeline.py
```diff
@@ -74,21 +73,31 @@ def extract_signal_quality_features(
...
 
-    acc_windowed_colnames = [DataColumns.TIME] + acc_config.accelerometer_colnames
-    acc_windowed = tabulate_windows(
-        df=df_acc,
-        columns=acc_windowed_colnames,
-        window_length_s=acc_config.window_length_s,
-        window_step_length_s=acc_config.window_step_length_s,
-        fs=acc_config.sampling_frequency,
-    )
+    df_features = pd.DataFrame(start_time_ppg, columns=[DataColumns.TIME])
```
Erikpostt: The `ppg_colname` is an attribute of the `config` object, perhaps you could do the same for the column indicating time? So, instead of using `DataColumns.TIME`, use something like `config.time_colname`. You can check [here](https://github.com/biomarkersParkinson/paradigma/blob/1039e037c3101430ff314dc05dd91faa891ad08e/src/paradigma/pipelines/gait_pipeline.py#L40) how I did it before for the gait pipeline.

Edit: the reason is that users can specify how they want to denote their `time` column. So, it need not be done for other columns, such as feature names.

src/paradigma/preprocessing.py
```diff
@@ -264,14 +264,18 @@ def preprocess_imu_data(
 
 def preprocess_ppg_data(
```
Erikpostt, line 265: This function returns `Tuple[pd.DataFrame, pd.DataFrame]`, but in your docstring you state that it may also return `None` if preprocessed accelerometer data is not provided. Did you run `poetry run pytype .` by any chance locally to test whether the docstrings are correct?

Erikpostt (changes requested):
Good work, I've added some additional suggestions. You could also add Copilot as reviewer, it may also find some easily overlooked issues.

docs/tutorials/_static/pulse_rate_analysis.md
KarsVeldkamp: Agree, I have adjusted it now to a variant of your second suggestion

docs/tutorials/_static/pulse_rate_analysis.md
KarsVeldkamp: I did already adjust this in my notebook tutorial but the markdown is not updated somehow?

docs/tutorials/_static/pulse_rate_analysis.md
KarsVeldkamp: Did this in the tutorial but same issue as above

src/paradigma/pipelines/pulse_rate_pipeline.py
KarsVeldkamp: Allright, I have adjusted this now for all DataColumns.TIME and DataColumns.PPG.

KarsVeldkamp: Should we remove them from constants.py too? Or do we use them somewhere in our analyses?

src/paradigma/preprocessing.py
KarsVeldkamp: I did run poetry run pytype . and they all passed. But i have now slightly adjusted the docstring and the return as optional for the accelerometer to make it clear

src/paradigma/preprocessing.py
```diff
@@ -264,14 +264,18 @@ def preprocess_imu_data(
...
     ppg_config: PPGConfig,
-    imu_config: IMUConfig,
-    start_time_ppg: str,
-    start_time_imu: str,
-) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    start_time_ppg: str | None = None,
+    df_acc: pd.DataFrame | None = None,
+    imu_config: IMUConfig | None = None,
+    start_time_imu: str | None = None,
+) -> Tuple[pd.DataFrame, Optional[pd.DataFrame]]:
```
Erikpostt:
```suggestion
) -> Tuple[pd.DataFrame, pd.DataFrame | None]:
```

Erikpostt: No need to import `Optional` here, it has been replaced by using `|`. You also specify this in the docstring below.

Erikpostt (approved):
Looking good, *almost* ready to merge. I have added one suggestiong for replacing `Optional`.

PR #222, opened 2025-09-25
Adding pre-commit hooks

Erikpostt:
## Description
This is a major change to developers' way of working and automates a lot of code cleanup, alleviating reviewers from this task. 

1. Our way of working was: change notebooks --> run notebook cells in correct order --> convert notebooks to markdown --> create docs --> inspect docs. With the proposed changes it is: `poetry run make-docs` (and to inspect, `poetry run serve-docs`).
2. Order of imports, unused variables, empty lines, etc. make code messy and unnecessarily complex. This can be automated.

## Changes
* Added pre-commit hooks (pre-commit-config.yaml) that run each time a change is committed.
* * nbstripout: cleans up notebooks and removes metadata (solving the kernel issue)
* * black, isort, ruff: cleans up Python code

* Created directory `scripts` containing two files: (1) `build_docs.py` and (2) `serve_docs.py`
* Added script references to `pyproject.toml` such that docs can be build with `poetry run build-docs` and served locally using `poetry run serve-docs`

* Updated `CONTRIBUTING.md` to include Poetry scripts and pre-commit hooks.

## Additional Information
All files were cleaned using pre-commit hooks, resulting in this large number of file changes. 

## Checklist

- [x] Tests passed
- [x] Documentation updated


Erikpostt:
@nienketimmermans apologies in advance that the burden to check this PR lies on you. We can discuss this if needed later.

docs/tutorials/_static/gait_analysis.md
```diff
@@ -1396,25 +1396,25 @@ pprint(arm_swing_aggregations, sort_dicts=False)
 ```
 
     {'0_10': {'duration_s': 341.25,
-              'median_range_of_motion': 15.10889561336818,
-              '95p_range_of_motion': 52.27598270995837,
-              'median_peak_velocity': 74.63996138215876,
-              '95p_peak_velocity': 230.71389576502395},
+              'median_range_of_motion': 15.108895613368176,
```
nienketimmermans, line 1399: The aggregated measures have all very slightly changed. Although this is not really and issue, I was wondering what caused this

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -1159,14 +1161,13 @@ from paradigma.pipelines.pulse_rate_pipeline import aggregate_pulse_rate
...
 )
 
 pprint.pprint(df_pr_agg)
 ```
 
-    {'metadata': {'nr_pr_est': 806},
-     'pr_aggregates': {'99p_pulse_rate': 87.65865011636926,
-                       'mode_pulse_rate': 81.25613346418058}}
-    
+    {'metadata': {'nr_pr_est': 8660},
```
nienketimmermans, line 1171: The pulse rate values have changed too, is this because the example data has changed?

nienketimmermans (commented):
This will be helpful! I did notice some changes to the aggregated output of the gait and pulse rate analysis and also to the example data, could you check why that happened?

docs/tutorials/_static/gait_analysis.md
Erikpostt: I'm actually not sure. This could be due to Python upgraded from 3.12 to 3.17, but I wouldn't expect that to happen at 1e-8. It also only happens at quantification, which should rule this out. Since it's only at the nth decimal, I wouldnt't worry too much about it. The change at pulse rate is more substantial and requires investigation.

docs/tutorials/_static/pulse_rate_analysis.md
Erikpostt: This is a substantial change, but it can be explained. Previously, the penultimate cell was not run (the cell running two segments). Therefore, the quantification was based on a single segment, while not in is for two segments. Since the pre-commit hooks automatically run all cells, this change occurs. I would recommend keeping it this way for now, and perhaps change the notebook later if this result is undesired. 

By any means, this change should require approval from @KarsVeldkamp.

KarsVeldkamp: The higher number of estimates are probably because segment 2 contains a full night of data while segment 1 only has daytime data.

nienketimmermans (approved):

Issue #216, opened 2025-09-17
Add parameter tolerance to preprocess_imu_data


Erikpostt:
**Is your feature request related to a problem? Please describe.**
As a user, I want to set the contiguous timestamp tolerance at a higher level than at the `resample_data` level. At this moment, if I want to change the tolerance, the only way is to copy the `preprocess_imu_data` function and change the parameter in `resample_data`, which is inconvenient.

**Describe the solution you'd like**
Add the hyperparameter `tolerance` to the functions `preprocess_imu_data` and `preprocess_ppg_data`. Also add this tolerance to `config`.

PR #223, opened 2025-09-26
Add tolerance for contiguous data

nienketimmermans:
## Description

Solves #216 

## Changes

The hyperparameter `tolerance` for contigous timestamps is now added to `config` (with default None), and given as argument to the `resample_data` function in the functions `preprocess_imu_data` and `preprocess_ppg_data`. This enables users to adjust the tolerance.  

## Additional Information

Only the last three commits are relevant.

## Checklist

- [x] Tests passed
- [x] Documentation updated


Erikpostt (changes requested):
Nice changes. Couple of recommendations:
1. `preprocess_imu_data` is run in the tutorials, such as for [tremor](https://github.com/biomarkersParkinson/paradigma/blob/01fcc26b4b1180f8abfd89135e3331f926edc173/docs/tutorials/_static/tremor_analysis.md). Make sure the `tolerance` hyperparameter is added to the tutorials, and explain to users how this can be set. Please also re-run the notebooks / create docs.
2. The function `resample_data` inside `preprocess_imu_data` has default value `None`, which is set to `3 * 1 / sampling_frequency`. I'd suggest setting the default value to `IMUConfig().tolerance` now that you added it to the config. This way, we only have to change it in one location (in `config.py`), instead of both at `config.py` and in this function. Also, I would set it to `3 * 1 / sampling_frequency` inside config.
3. In `preprocess_ppg_data`,  `resample_data` is called twice, once for IMU and once for PPG. However, for PPG is does not have a tolerance parameter. Since it also defaults to `3 * 1 / sampling_frequency`, perhaps we should set it inside `config.py` as well?

nienketimmermans:
Good suggestions, I incorporated them now. I also noticed that the resampling frequency for ppg data was still set to the sampling frequency in `preprocess_ppg_data`, so I changed this too.

docs/tutorials/gait_analysis.ipynb
```diff
@@ -249,10 +249,10 @@
    "metadata": {},
    "source": [
     "## Step 1: Preprocess data\n",
-    "The single function `preprocess_imu_data` in the cell below runs all necessary preprocessing steps. It requires the loaded dataframe, a configuration object `config` specifying parameters used for preprocessing, and a selection of sensors. For the sensors, options include `'accelerometer'`, `'gyroscope'`, or `'both'`.\n",
+    "The single function `preprocess_imu_data` in the cell below runs all necessary preprocessing steps. It requires the loaded dataframe, a configuration object `config` specifying parameters used for preprocessing, and a selection of sensors. For the sensors, options include `'accelerometer'`, `'gyroscope'`, or `'both'`.  If the difference between timestamps is larger than a specified tolerance (`config.tolerance`), it will return an error that the timestamps are not contigous.\n",
```
Erikpostt: Since `tolerance = 3 * 1 / sampling_frequency`, perhaps it's best to specify that the tolerance has _seconds_ as units.

docs/tutorials/gait_analysis.ipynb
```diff
@@ -249,10 +249,10 @@
...
     "\n",
     "The function `preprocess_imu_data` processes the data as follows:\n",
-    "1. Resample the data to ensure uniformly distributed sampling rate\n",
+    "1. Resample the data to ensure uniformly distributed sampling rate.\n",
```
Erikpostt, line 255: If you add an interpunction here, why not in the next enumeration? :)

docs/tutorials/gait_analysis.ipynb
Erikpostt: Small typo in final line: `contigous` -> `contiguous`.

Erikpostt: Perhaps we could add here that if the user still wants to process this data, they can create segments from discontiguous samples using `create_segments` and then process the segments consecutively as in one of the final cells of each notebooks using a for-loop. What do you think?

docs/tutorials/pulse_rate_analysis.ipynb
```diff
@@ -324,7 +324,7 @@
...
+    "The first step after loading the data is preprocessing using the [preprocess_ppg_data](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/preprocessing.py#:~:text=preprocess_ppg_data). This begins by isolating segments containing both PPG and IMU data, discarding portions where one modality (e.g., PPG) extends beyond the other, such as when the PPG recording is longer than the accelerometer data. This functionality requires the starting times (`metadata_time_ppg.start_iso8601` and `metadata_time_imu.start_iso8601`) in iso8601 format as inputs. After this step, the preprocess_ppg_data function resamples the PPG and accelerometer data to uniformly distributed timestamps, addressing the fixed but non-uniform sampling rates of the sensors. If the difference between timestamps is larger than a specified tolerance (`config.tolerance`), it will return an error that the timestamps are not contigous.After this, a bandpass Butterworth filter (4th-order, bandpass frequencies: 0.4--3.5 Hz) is applied to the PPG signal, while a high-pass Butterworth filter (4th-order, cut-off frequency: 0.2 Hz) is applied to the accelerometer data.\n",
```
Erikpostt:
```suggestion
    "The first step after loading the data is preprocessing using the [preprocess_ppg_data](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/preprocessing.py#:~:text=preprocess_ppg_data). This begins by isolating segments containing both PPG and IMU data, discarding portions where one modality (e.g., PPG) extends beyond the other, such as when the PPG recording is longer than the accelerometer data. This functionality requires the starting times (`metadata_time_ppg.start_iso8601` and `metadata_time_imu.start_iso8601`) in iso8601 format as inputs. After this step, the preprocess_ppg_data function resamples the PPG and accelerometer data to uniformly distributed timestamps, addressing the fixed but non-uniform sampling rates of the sensors. If the difference between timestamps is larger than a specified tolerance (`config.tolerance`), it will return an error that the timestamps are not contiguous. After this, a bandpass Butterworth filter (4th-order, bandpass frequencies: 0.4--3.5 Hz) is applied to the PPG signal, while a high-pass Butterworth filter (4th-order, cut-off frequency: 0.2 Hz) is applied to the accelerometer data.\n",
```

Erikpostt: Small typos.

docs/tutorials/pulse_rate_analysis.ipynb
```diff
@@ -338,6 +338,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
+      "The tolerance for checking contiguous timestamps is set to 0.100 seconds for PPG and 0.030 seconds for Accelerometer data.\n",
```
Erikpostt:
```suggestion
      "The tolerance for checking contiguous timestamps is set to 0.100 seconds for PPG data and 0.030 seconds for accelerometer data.\n",
```

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -255,7 +255,7 @@
...
+    "IMU sensors collect data at a fixed sampling frequency, but the sampling rate is not uniform, causing variation in time differences between timestamps. The [preprocess_imu_data](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/preprocessing.py#:~:text=preprocess_imu_data) function therefore resamples the timestamps to be uniformly distributed, and then interpolates IMU values at these new timestamps using the original timestamps and corresponding IMU values. If the difference between timestamps is larger than a specified tolerance (`config.tolerance`), it will return an error that the timestamps are not contigous. By setting `sensor` to 'gyroscope', only gyroscope data is preprocessed and the accelerometer data is removed from the dataframe. Also a `watch_side` should be provided, although for the tremor analysis it does not matter whether this is the correct side since the tremor features are not influenced by the gyroscope axes orientation."
```
Erikpostt:
```suggestion
    "IMU sensors collect data at a fixed sampling frequency, but the sampling rate is not uniform, causing variation in time differences between timestamps. The [preprocess_imu_data](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/preprocessing.py#:~:text=preprocess_imu_data) function therefore resamples the timestamps to be uniformly distributed, and then interpolates IMU values at these new timestamps using the original timestamps and corresponding IMU values. If the difference between timestamps is larger than a specified tolerance (`config.tolerance`), it will return an error that the timestamps are not contiguous. By setting `sensor` to 'gyroscope', only gyroscope data is preprocessed and the accelerometer data is removed from the dataframe. Also a `watch_side` should be provided, although for the tremor analysis it does not matter whether this is the correct side since the tremor features are not influenced by the gyroscope axes orientation."
```

Erikpostt (changes requested):
Several typos and minor suggestions; I think a final sweep should do the trick.

docs/tutorials/gait_analysis.ipynb
nienketimmermans: This was not done on purpose haha, but I have added it to the next enumeration as well now.

docs/tutorials/gait_analysis.ipynb
nienketimmermans: Yes, good idea

nienketimmermans:
Thank you for noticing the typos, they should all be fixed now.

docs/tutorials/gait_analysis.ipynb
```diff
@@ -249,11 +249,11 @@
    "metadata": {},
    "source": [
     "## Step 1: Preprocess data\n",
-    "The single function `preprocess_imu_data` in the cell below runs all necessary preprocessing steps. It requires the loaded dataframe, a configuration object `config` specifying parameters used for preprocessing, and a selection of sensors. For the sensors, options include `'accelerometer'`, `'gyroscope'`, or `'both'`.\n",
+    "The single function `preprocess_imu_data` in the cell below runs all necessary preprocessing steps. It requires the loaded dataframe, a configuration object `config` specifying parameters used for preprocessing, and a selection of sensors. For the sensors, options include `'accelerometer'`, `'gyroscope'`, or `'both'`.  If the difference between timestamps is larger than a specified tolerance (`config.tolerance`, in seconds), it will return an error that the timestamps are not contiguous. If you still want to process the data in this case, you can create segments from discontiguous samples using the function `create_segments` and analyze these segments consecutively as shown in [here](#multiple_segments_cell).\n",
```
Erikpostt:
```suggestion
    "The single function `preprocess_imu_data` in the cell below runs all necessary preprocessing steps. It requires the loaded dataframe, a configuration object `config` specifying parameters used for preprocessing, and a selection of sensors. For the sensors, options include `'accelerometer'`, `'gyroscope'`, or `'both'`.  If the difference between timestamps is larger than a specified tolerance (`config.tolerance`, in seconds), it will return an error that the timestamps are not contiguous. If you still want to process the data in this case, you can create segments from discontiguous samples using the function [`create_segments`](https://github.com/biomarkersParkinson/paradigma/blob/main/src/paradigma/segmenting.py) and analyze these segments consecutively as shown [here](#multiple_segments_cell).\n",
```

Erikpostt (approved):
I left a minor suggestion to link to the location of `create_segments`, but otherwise ready to merge.

Issue #224, opened 2025-09-30
Add flexibility of column names for tremor analysis


nienketimmermans:
**Is your feature request related to a problem? Please describe.**
Currently, columns generated as output in the tremor analysis (such as `pred_tremor_checked`) have fixed names.

**Describe the solution you'd like**
Users should be able to specify their own column names for tremor output, by defining them in the `config` file.

PR #241, opened 2025-12-17
Add flexibility for tremor columns

nienketimmermans:
## Description

Solves #224 by adding the ability for users to specify the tremor column names.

## Changes

- I've added column names used in the tremor analysis in `constants.py` and changed the tutorials and `tremor_analysis.py` accordingly. 
- I also fixed a typo (interally -> internally) in the gait, pulse rate and tremor tutorials. 
- In the tremor tutorial the `step` parameter in `TremorConfig` is not necessary, so I removed it there. It is only necessary when storing all the features in tsdf, for example. 
- I had some issues with installing jupyter-lab widgets, so I changed the `pyproject.toml` file to make sure version 3.0.15 instead of 3.0.16 is installed.

## Checklist

- [x] Tests passed
- [x] Documentation updated


docs/tutorials/_static/gait_analysis.md
```diff
@@ -1226,8 +1226,8 @@ quantified_arm_swing.loc[quantified_arm_swing['segment_nr'] == 1]
     Metadata of the first gait segment:
     {'duration_filtered_segment_s': 9.0,
      'duration_unfiltered_segment_s': 9.0,
-     'end_time_s': 2230.74,
-     'start_time_s': 2221.75}
+     'end_time_s': np.float64(2230.74),
+     'start_time_s': np.float64(2221.75)}
```
KarsVeldkamp: @nienketimmermans @Erikpostt Is this what we intend to print here? I see this is the case for gait, tremor and pulse rate aggregations

nienketimmermans: This was changed by the pre-commit hooks

KarsVeldkamp: But why did this change in this PR? Since in main this is not the case right? I am just wondering...

KarsVeldkamp (approved):
Looking good ;)

docs/tutorials/_static/data_preparation.md
```diff
@@ -9,7 +9,7 @@ ParaDigMa requires the sensor data to be of a specific format. This tutorial pro
 The final dataframe should be resampled to 100 Hz, have the correct units for the sensor columns, and the correct format for the time column. Also note that the _gait_ pipeline expects a specific orientation of sensor axes, as explained in [Coordinate system](../guides/coordinate_system.md).
 
 ## Load data
-This example uses data of the [Personalized Parkinson Project](https://pubmed.ncbi.nlm.nih.gov/31315608/), which is stored in Time Series Data Format (TSDF). Inertial Measurements Units (IMU) and photoplethysmography (PPG) data are sampled at a different sampling frequency and therefore stored separately. Note that ParaDigMa works independent of data storage format; it only requires a `pandas` dataframe as input.
+This example uses data of the [Personalized Parkinson Project](https://pubmed.ncbi.nlm.nih.gov/31315608/), which is stored in Time Series Data Format (`TSDF`). Inertial Measurements Units (IMU) and photoplethysmography (PPG) data are sampled at a different sampling frequency and therefore stored separately. Note that ParaDigMa works independent of data storage format; it only requires a `pandas.DataFrame` as input.
```
Erikpostt, line 12: Any particular reason for rephrasing `pandas.DataFrame` to `pandas` dataframe?

docs/tutorials/_static/gait_analysis.md
Erikpostt: Yes I am also curious, in particular because not all `float` parameters have this property. Perhaps change the floats to `float` format instead of `numpy.float`.

docs/tutorials/_static/gait_analysis.md
```diff
@@ -1536,25 +1536,25 @@ pprint(arm_swing_aggregations, sort_dicts=False)
...
+                '95p_peak_velocity': np.float64(217.64806342438817)},
      '0_inf': {'duration_s': 2307.75,
-               'median_range_of_motion': 24.071313521090392,
-               '95p_range_of_motion': 43.06891252479754,
-               'median_peak_velocity': 120.43812492382041,
-               '95p_peak_velocity': 215.7685538864721}}
+               'median_range_of_motion': np.float64(24.07131352109043),
+               '95p_range_of_motion': np.float64(43.06891252479739),
+               'median_peak_velocity': np.float64(120.43812492382015),
+               '95p_peak_velocity': np.float64(215.76855388647215)}}
```
Erikpostt: Same here.

docs/tutorials/_static/pulse_rate_analysis.md
```diff
@@ -1256,5 +1256,5 @@ pprint.pprint(df_pr_agg)
 ```
 
     {'metadata': {'nr_pr_est': 8660},
-     'pr_aggregates': {'99p_pulse_rate': 85.77263444520081,
-                       'mode_pulse_rate': 63.59175662414131}}
+     'pr_aggregates': {'99p_pulse_rate': np.float64(85.77263444520081),
+                       'mode_pulse_rate': np.float64(63.59175662414131)}}
```
Erikpostt: And here.

docs/tutorials/_static/tremor_analysis.md
```diff
@@ -792,14 +793,14 @@ metadata_values_store = tsdf.TSDFMetadata(
 # Select the columns to be saved
 metadata_time_store.channels = [config.time_colname]
 metadata_values_store.channels = [
-    'tremor_power', 'pred_tremor_proba', 'pred_tremor_logreg',
-    'pred_arm_at_rest', 'pred_tremor_checked'
+    DataColumns.TREMOR_POWER, DataColumns.PRED_TREMOR_PROBA, DataColumns.PRED_TREMOR_LOGREG,
```
Erikpostt, line 796: Suggestions: change `PRED_TREMOR_LOGREG` to `PRED_TREMOR`, unless you want to specify that the model is a logistic regression.

docs/tutorials/_static/tremor_analysis.md
```diff
@@ -792,14 +793,14 @@ metadata_values_store = tsdf.TSDFMetadata(
...
+    DataColumns.PRED_ARM_AT_REST, DataColumns.PRED_TREMOR_CHECKED
 ]
 
 # Set the units
 metadata_time_store.units = ['Relative seconds']
 metadata_values_store.units = [
-    'Unitless', 'Unitless', 'Unitless', 'Unitless', 'Unitless'
+    DataUnits.NONE, DataUnits.NONE, DataUnits.NONE, DataUnits.NONE, DataUnits.NONE
```
Erikpostt: Isn't `DataColumns.TREMOR_POWER` in `g^2/Hz`?

docs/tutorials/_static/tremor_analysis.md
```diff
@@ -1195,10 +1196,10 @@ pprint.pprint(d_tremor_aggregates)
...
     Before aggregation we select data collected between 08:00 and 22:00. We also select days with at least 10 hours of data.
     The following tremor power aggregates are derived: ['mode_binned', 'median', '90p'].
-    {'aggregated_tremor_measures': {'90p_tremor_power': 1.3259483071516063,
-                                    'median_tremor_power': 0.5143985314908104,
-                                    'modal_tremor_power': 0.3,
-                                    'perc_windows_tremor': 19.386769676484793},
+    {'aggregated_tremor_measures': {'90p_tremor_power': np.float64(1.3259483071516063),
+                                    'median_tremor_power': np.float64(0.5143985314908104),
+                                    'modal_tremor_power': np.float64(0.3),
+                                    'perc_windows_tremor': np.float64(19.386769676484793)},
```
Erikpostt: And here.

docs/tutorials/tremor_analysis.ipynb
```diff
@@ -219,14 +220,14 @@
...
-    "    'pred_arm_at_rest', 'pred_tremor_checked'\n",
+    "    DataColumns.TREMOR_POWER, DataColumns.PRED_TREMOR_PROBA, DataColumns.PRED_TREMOR_LOGREG, \n",
+    "    DataColumns.PRED_ARM_AT_REST, DataColumns.PRED_TREMOR_CHECKED\n",
     "]\n",
     "\n",
     "# Set the units\n",
     "metadata_time_store.units = ['Relative seconds']\n",
     "metadata_values_store.units = [\n",
-    "    'Unitless', 'Unitless', 'Unitless', 'Unitless', 'Unitless'\n",
+    "    DataUnits.NONE, DataUnits.NONE, DataUnits.NONE, DataUnits.NONE, DataUnits.NONE\n",
```
Erikpostt: Also check units here for `DataColumns.TREMOR_POWER`.

pyproject.toml
```diff
@@ -48,6 +48,7 @@ jupyter-client = "^8.6.3"
 nbstripout = "^0.8.1"
 nb-clean = "^4.0.1"
 jupyter = "^1.1.1"
+jupyterlab-widgets = "3.0.15"
```
Erikpostt: I would suggest changing it to `<=3.0.15` to add some flexibility (now you're forcing a version that may be incompatible with other dependencies).

src/paradigma/config.py
```diff
@@ -266,11 +266,16 @@ def __init__(self, step: str | None = None) -> None:
...
             for mfcc_coef in range(1, self.n_coefficients_mfcc + 1):
-                self.d_channels_values[f"mfcc_{mfcc_coef}"] = "unitless"
+                self.d_channels_values[f"mfcc_{mfcc_coef}"] = DataUnits.NONE
+
+            self.d_channels_values[DataColumns.FREQ_PEAK] = DataUnits.FREQUENCY
+            self.d_channels_values[DataColumns.BELOW_TREMOR_POWER] = (
+                DataUnits.POWER_SPECTRAL_DENSITY_ROT
+            )
+            self.d_channels_values[DataColumns.TREMOR_POWER] = (
+                DataUnits.POWER_SPECTRAL_DENSITY_ROT
```
Erikpostt: Is `DataUnits.POWER_SPECTRAL_DENSITY_ROT` not `(deg/s)^2/Hz` instead of `(deg/s)^2`?

Erikpostt (commented):
Looks good overall, I've made several suggestions.

docs/tutorials/_static/data_preparation.md
nienketimmermans: By the pre-commit hooks it was changed from `pandas` dataframe to `pandas.DataFrame`

docs/tutorials/_static/tremor_analysis.md
nienketimmermans: The final tremor prediction (`PRED_TREMOR_CHECKED`) depends on the output of the logistic regression classifier and some extra checks, so I think it is better to name the output of the classifier `PRED_TREMOR_LOGREG` to avoid confusion about the tremor prediction

docs/tutorials/_static/tremor_analysis.md
nienketimmermans: No because tremor power is calculated as area under the curve of the power spectral density

pyproject.toml
nienketimmermans: Good point

src/paradigma/config.py
nienketimmermans: I should have named it `POWER_ROTATION` instead of `POWER_SPECTRAL_DENSITY_ROT`, so I changed it now

docs/tutorials/tremor_analysis.ipynb
nienketimmermans: Yes, good point

docs/tutorials/_static/gait_analysis.md
nienketimmermans: @Erikpostt Do you mean to change this in the md files?

Erikpostt: No, what I mean is to change the variables in the text from a `numpy.float` object to a `float` object (`float` is an inherent object of Python). I just checked why this is changed suddenly when using `pre-commit`, and it has to do with the `numpy` upgrade to `2.X`. Now, scalar types are set explicitly. So, some parameters are of type `float`, while others are of type `numpy.float`. This is not an issue per se, it's just less "clean". Couple of options:
1. Use a function (in `util.py`) to convert to `float`, such as:

```Python
def to_py(x):
    return x.item() if isinstance(x, np.generic) else x

pprint({k: to_py(v) for k, v in d.items()})
```

2. Keep it as it is.
3. Change from `pprint` to `print` (which uses `str()` instead of `repr()`, the latter causing the dtype showing). This does require some indentation for the dictionary, for example by:

```Python
print(json.dumps(dict, indent=2))
```

So don't change it manually in the `.md` files, try to fix it in the `.ipynb` files.

docs/tutorials/_static/data_preparation.md
Erikpostt: I read it the other way around, my bad.

nienketimmermans:
@Erikpostt I changed the `.ipynb` files by changing from `pprint` to `print`

Erikpostt:
> @Erikpostt I changed the `.ipynb` files by changing from `pprint` to `print`

Changes look good, no further open issues from my side.

PR #278, opened 2026-03-26
Change arm swing metadata to datetime

Erikpostt:
## Description

Add datetime metadata to arm swing quantification results. This enhancement enables absolute timestamp tracking for each arm swing segment, providing more meaningful temporal context for analysis while maintaining full backward compatibility.

## Changes

- **src/paradigma/load.py**: Added datetime extraction from TSDF, Empatica, Axivity, and prepared DataFrame formats. Returns 3-tuple: `(file_key, DataFrame, datetime|None)`
- **src/paradigma/orchestrator.py**: Updated to unpack 3-tuple and pass `start_dt` to `run_gait_pipeline()`
- **src/paradigma/pipelines/gait_pipeline.py**: Added `start_dt` parameter to conditionally include datetime in metadata
  - With datetime: metadata includes `start_dt` and `end_dt` (ISO 8601 format)
  - Without datetime: metadata uses relative time offsets (`start_s`, `end_s`)
  - Result DataFrames remain identical in both cases
- **tests/test_gait_analysis.py**: Updated mock function to accept `start_dt` parameter

### Key Differences in Metadata Output

**With datetime**: Segments include absolute timestamps
- `'start_dt': '2019-08-20T11:03:39Z'`
- `'end_dt': '2019-08-20T11:03:57.990000Z'`

**Without datetime**: Segments use relative offsets
- `'start_s': 1463.0`
- `'end_s': 1481.99`

### Testing
- All 15 tests pass
- Verified with 3.4M row TSDF dataset (9.5 hours)
- Fully backward compatible

## Checklist

- [x] Tests passed
- [x] Documentation updated


Erikpostt:
@nienketimmermans this should solve the discrepancy between the old quantification files (as on Snellius) and the new ones. Currently, in Paradigma, no datetime information is stored, and we (and other users too) need this.

src/paradigma/load.py
```diff
@@ -367,26 +371,46 @@ def load_single_data_file(
...
             # For TSDF, load based on .meta file and infer prefix
             if file_path.suffix.lower() == ".json":
                 prefix = file_path.stem.replace("_meta", "")
-                df, _, _ = load_tsdf_data(file_path.parent, prefix)
-                return prefix, df
+                df, time_meta, _ = load_tsdf_data(file_path.parent, prefix)
+                # Extract start_iso8601 from TSDF metadata
+                if hasattr(time_meta, "start_iso8601"):
+                    start_iso8601 = time_meta.start_iso8601
+                    start_dt = datetime.fromisoformat(start_iso8601.rstrip("Z"))
```
nienketimmermans: Why do you remove the timezone info here?

nienketimmermans (approved):
This would indeed be useful! I have a question about the handling of timezone information (see below), but otherwise ready to merge

src/paradigma/load.py
Erikpostt: This is because `datetime.fromisoformat` is unable to take the "Z" for UTC into account. I therefore propose we remove it now, and we can add a feature taking into account timezone information later.

nienketimmermans: I read somewhere that as of python version 3.11 this should be possible (see https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat), and python ^3.11 is a requirement right?

Erikpostt: You are absolutely right, I have changed `load.py` and `gait_pipeline.py` accordingly.

Erikpostt:
@nienketimmermans can you check whether you agree with the changes?

nienketimmermans:
@Erikpostt yes, this looks good now!

KarsVeldkamp (approved):

PR #281, opened 2026-04-03
Fix fs dependency tremor pipeline

nienketimmermans:
## Description

The tremor pipeline worked well for gyroscope signals sampled at 100Hz, but did not for signals sampled at lower frequencies. 
With these changes the pipeline works well for sampling frequencies as low as 64Hz. 

## Changes

- Computation of the spectrogram was adjusted to make it independent of the sampling frequency used.
- Also a bug related to storing intermediate results was fixed.

## Additional Information

The changes were tested for one week of data of 8 patients. Weekly tremor measures were similar for 100Hz and 64Hz downsampled signals.  

## Checklist

- [x] Tests passed
- [x] Documentation updated

### DIFF
--- a/src/paradigma/pipelines/tremor_pipeline.py
+++ b/src/paradigma/pipelines/tremor_pipeline.py
@@ -339,9 +339,9 @@ def extract_spectral_domain_features(data: np.ndarray, config) -> pd.DataFrame:
 
     # Compute the power spectral density
     segment_length_n = sampling_frequency * segment_length_psd_s
-    overlap_n = int(segment_length_n * overlap_fraction)
+    overlap_n = int(np.round(segment_length_n * overlap_fraction))
     window = signal.get_window(window_type, segment_length_n, fftbins=False)
-    nfft = int(sampling_frequency / spectral_resolution_psd)
+    nfft = int(np.round(sampling_frequency / spectral_resolution_psd))
 
     freqs, psd = signal.welch(
         x=data,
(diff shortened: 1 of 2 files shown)


src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -356,23 +357,27 @@ def extract_spectral_domain_features(data: np.ndarray, config) -> pd.DataFrame:
...
         axis=1,
     )
 
     # Compute total power in the PSD and the total spectrogram (summed over
     # the three axes)
     total_psd = compute_total_power(psd)
-    total_spectrogram = np.sum(np.abs(stft_result) * sampling_frequency, axis=2)
+    total_spectrogram = np.sum(
+        np.abs(stft_result) * 100, axis=2
+    )  # scaling factor of 100 to match with previous Matlab code
```
Erikpostt: "to match with previous Matlab code" is slightly ambiguous, perhaps you can clarify that the original model was developed using a 100 Hz sample if that's relevant?

src/paradigma/pipelines/tremor_pipeline.py
```diff
@@ -356,23 +357,27 @@ def extract_spectral_domain_features(data: np.ndarray, config) -> pd.DataFrame:
 
     # Compute the spectrogram
     segment_length_n = sampling_frequency * segment_length_spectrogram_s
-    overlap_n = segment_length_n * overlap_fraction
+    overlap_n = int(segment_length_n * overlap_fraction)
```
Erikpostt: `int` truncates the floating point number to integer format. You may want to consider using `int(np.round)` instead to ensure 2.8 becomes 3 instead of 2 (instead this is what you intended).

Erikpostt:
Looks good, I've made two small suggestions but otherwise ready to merge.

Erikpostt (approved):

src/paradigma/pipelines/tremor_pipeline.py
nienketimmermans: Good point, I will add that

PR #285, opened 2026-04-16
feat: implement adaptive frequency detection and auto-parameter adjustment

Erikpostt:
## Description
This PR implements the foundation for adaptive frequency detection and automatic parameter adjustment across ParaDigMa pipelines. This is based on making ParaDigMa work for lower sampling frequencies, where we noticed that upsampling is inferior to keeping the same frequency but adjusting frequency-based features accordingly.

The changes should not break / change much.

## Changes
- **Frequency Detection**: Added `detect_sampling_frequency()` utility with median-based detection
- **Config System**: Implemented `sampling_frequency` property setter with auto-update of frequency-dependent parameters
- **Orchestrator Integration**: Added automatic frequency detection after data preparation
- **Preprocessing**: Made `resampling_frequency` optional (maintains current frequency by default)
- **Documentation**: Updated guides explaining frequency flexibility

## Key Features
✅ Backward compatible with existing code


Erikpostt:
Couple of notes:

* The gait output is intended to change: if you look at the `duration_s` of 18 seconds, it should belong in `0_20` instead of `20_inf`, so this fixed an earlier error.
* Gait pipeline adds an additional `1 / fs` to `end_s`, because `time` represents the start of a window, so `end_s` represented the start time of the final window, which should be the end time of the final window.
* Be critical about how we pose that our pipeline works >= 50 Hz, I'm not sure whether I phrased it honestly and covered the necessary details right.
* There were some auto-generated files created by the agent that should be removed.

nienketimmermans (commented):
If I understand it correctly, the resampling frequency is set to 100 Hz if it is not provided, but this will result in upsampling of the data if the sampling frequency is below 100 Hz right? Shouldn't it be set to the auto-detected sampling frequency to ensure uniform sampling at that frequency?

Erikpostt:
> If I understand it correctly, the resampling frequency is set to 100 Hz if it is not provided, but this will result in upsampling of the data if the sampling frequency is below 100 Hz right? Shouldn't it be set to the auto-detected sampling frequency to ensure uniform sampling at that frequency?

Yes, good catch! I think I've changed it correctly now. I've additionally removed resampling from data preparation (in `run_paradigma` with `skip_preparation=False`), which became redundant.

Note that I've now added segmentation to data preparation as this functionality was called inside the `resample_data` function, which was now removed.

docs/guides/sensor_requirements.md
```diff
@@ -19,15 +23,15 @@ All pipelines require data from a wrist-worn sensor with:
...
+| **Accelerometer** | Sampling rate ≥ 50 Hz<br>Range ≥ ± 4 g |
+| **Gyroscope** | Sampling rate ≥ 50 Hz<br>Range ≥ ± 1000 degrees/sec |
 
 #### Physical Units
 
 - **Accelerometer**: `g` (gravitational force)
 - **Gyroscope**: `deg/s` (degrees per second)
 
-> ParaDigMa has functionalities for converting [accelerometer](https://biomarkersparkinson.github.io/paradigma/autoapi/paradigma/util/index.html#paradigma.util.convert_units_accelerometer) and [gyroscope](https://biomarkersparkinson.github.io/paradigma/autoapi/paradigma/util/index.html#paradigma.util.convert_units_gyroscope) in other units (e.g., `m/s^2`, `rad/s`) to these standard units. This can also be setup when using `run_paradigma`.
+> ParaDigMa has functionalities for converting [accelerometer](https://biomarkersparkinson.github.io/paradigma/autoapi/paradigma/util/index.html#paradigma.util.convert_units_accelerometer) and [gyroscope](https://biomarkersparkinson.github.io/paradigma/autoapi/paradigma/util/index.html#paradigma.util.convert_units_gyroscope) in other units (e.g., `m/s^2`, `rad/s`) to these standard units. This can also be setup when using run_paradigma`.
```
KarsVeldkamp, line 34: Is the ` missing? Or should you also remove the latter one?

docs/guides/sensor_requirements.md
```diff
@@ -130,6 +134,6 @@ results = run_paradigma(
 ## Important Notes on Validation
 
 > [!NOTE]
-> The specifications above represent **minimally validated requirements**. For example, while ParaDigMa works with accelerometer and gyroscope data sampled at 50 Hz, the effect on processing accuracy has not been empirically validated.
+> The specifications above represent **validated requirements**. ParaDigMa has been tested and verified to work correctly with sampling frequencies as low as 50 Hz, with comparable accuracy to 100 Hz baseline.
```
KarsVeldkamp, line 137: If you state that it has been tested and verified to work correctly then the part after the comma becomes redundant right?

src/paradigma/config.py
```diff
@@ -211,6 +234,43 @@ def __init__(self, step, column_mapping: dict[str, str] | None = None) -> None:
...
                 )
 
+        # Update frequency-dependent parameters now that all attributes are initialized
+        self._update_frequency_dependent_params()
+
+    def _update_frequency_dependent_params(self) -> None:
+        """
+        Update frequency-dependent parameters when sampling_frequency changes.
+
+        Ensures that spectral bounds stay within Nyquist limit (fs/2).
```
KarsVeldkamp, line 244: Maybe a minor semantic thing but technically it is called Nyquist frequency right?

src/paradigma/prepare_data.py
```diff
@@ -378,15 +381,39 @@ def prepare_raw_data(
...
+            ),
+        )
+        df["data_segment_nr"] = segment_array
+
+        # Discard segments that are too short
+        df = discard_segments(
+            df=df,
+            segment_nr_colname="data_segment_nr",
+            min_segment_length_s=(
+                min_segment_length_s if min_segment_length_s is not None else 1.5
```
KarsVeldkamp, line 401: arbitrary?

src/paradigma/prepare_data.py
```diff
@@ -378,15 +381,39 @@ def prepare_raw_data(
...
+            ),
+            fs=100,  # Approximate fs for segment length calculation
+            format="timestamps",
+        )
+
+        n_segments = df["data_segment_nr"].nunique()
+        logger.info(f"Created {n_segments} segments")
+
+    # Deprecation notice for resampling_frequency parameter
+    if resampling_frequency != 100.0:
```
KarsVeldkamp, line 411: What if it is set at 100? Maybe not the most logical thing to do but then the user still expect to already have resampled data right? Or am I not understanding this correctly?

KarsVeldkamp (changes requested):
In general this is looking good, but I have a couple of questions before ready to merge

nienketimmermans:
@Erikpostt I have made adjustments to the tremor pipeline so that odd sampling frequencies work well too.

src/paradigma/preprocessing.py
nienketimmermans: Should we change this to sampling_frequency = int(1/current_dt)? The tremor pipeline currently relies on the sampling frequency being an integer, but not sure where to adjust this

Erikpostt:
I've made some additional changes after sitting down with Kars. In summary:

Documentation updates
* Removed manual `sampling_frequency` setting from examples
* Updated to reflect auto-detected sampling frequency during preprocessing
* Introduced `resampling_frequency` as an optional parameter for upsampling/downsampling

Config
* Implemented frequency-dependent parameter adaptation
* `sampling_frequency` now auto-detected instead of manually set
* Added `resampling_frequency` parameter for optional resampling (not recommended)

Pipelines updated accordingly

Preprocessing
* Moved frequency detection into preprocessing phase (previously in both data preparation and preprocessing)
* Cleaner separation of concerns

@nienketimmermans am I correct that `segment_length_n` should also be an integer? I've changed this to be sure. Also, I'm not sure whether the `round_detected_frequency` is what we wanted to do for sampling frequencies; I believe you informed me that uneven sampling frequencies should be fine too.

EDIT: this needs a bit more work and testing. Let's discuss this later.

Erikpostt:
@nienketimmermans if we do automatic resampling, the test data is actually resampled to 101 Hz (instead of 100 Hz) causing quite some changes. Could you check the markdown files and inspect these changes? Perhaps we should be more lenient on rounding (e.g., use only even sampling rates or something similar). What do you think? We can discuss on Monday if needed.

nienketimmermans (approved):
Ready to merge!

Erikpostt:
Thank you @nienketimmermans; only awaiting @KarsVeldkamp for approval (upon his return from leave).

KarsVeldkamp:
@Erikpostt The preprocess_ppg_data() function currently modifies the ppg_config object (and similar for imu_config in my own and your pipelines) by calling config._set_sampling_frequency_detected() internally. While this works, for me it makes the intent unclear to callers: it is not obvious that the config object will be mutated. I am not sure what if this is an issue at all but maybe we can discuss this further. This is my only issue with the current adjustments, so apart from this we are ready to merge

Erikpostt:
@KarsVeldkamp I've made some changes based on our discussion. Can you verify these changes and perhaps provide some additional suggestions if needed?

KarsVeldkamp (approved):
This looks good!

